PR #85, opened 2022-07-08
Remove Kubernetes 1.24 deprecated address argument for kube-controlle…

pysen:
With these changes the create-cluster command will also work for Kubernetes v1.24+


vitobotta:
Thanks, merging then. 👍

PR #110, opened 2022-10-31
Increase subnet mask to 24

cwilhelm:
Increase of subnet mask to 24 would gives the possibility to add vSwitch to the network. Currently the network and only existing subnet have the same subnet mask and therefor any other subnet has an overlap and is denied.
With a vSwitch I could probably connect a dedicated server to the cloud.


vitobotta:
Thanks for the PR! :)

PR #136, opened 2023-01-08
Update README.md

aleksasiriski:
Removed not working roadmap and made a few small adjustments to improve readability.


vitobotta (approved):

vitobotta:
Thanks a lot! I hadn't realized that the roadmap was not public after migrating to the new Github projects, but I was not really using it anyway. Thanks :)

aleksasiriski:
> Limitations:
> autoscaling is only supported when using Ubuntu or one of the other default images, not snapshots

Also, isn't this fixed with v0.1.5? Or it that working just for MicroOS currently?

vitobotta:
Oh yes it's fixed, so that line can be removed. Feel free to do another PR if you like :)

PR #137, opened 2023-01-08
QoL changes

aleksasiriski:
- Updated README
- Added latest tag


aleksasiriski:
I didn't test this "latest tag" but I guess it should *just work*? Help to not need to touch README on every release.

vitobotta:
Thanks! I will test it and let you know.

PR #153, opened 2023-01-27
Expand config path, allow for hetzner_token to come from a file

radekg:
Hey @vitobotta. Thank you very much for this project. I was able to start a k8s cluster in Hetzner in no time. While working with the tool, I came across a couple of problems:

- The main config path needed to be expanded. Passing `--config=~/path/to/file.yaml` would fail with an unhandled exception. I am providing a fix here.
- The second problem is more profound: I imagine one would like to commit the `cluster yaml config` file into a repository. However, there are better ideas than having the Hetzner token in the repository. This PR adds very rudimentary support for loading the token from a file. Basically, in the yaml file:

```yaml
---
hetzner_token: file://~/.hetzner/token
...
```

If the `hetzner_token` key value starts with `file://`, the program strips the `file://` prefix, expands the remaining value, checks if the file exists, reads the content, and uses that value as a token.

To prevent unexpected token exfiltration when a downloaded yaml file is used by the unsuspecting user and the path points to a sensitive file, the user has to explicitly allow the use of such paths by passing `--allow-token-from-file` flag.

Example:

```sh
hetzner-k3s create --config=~/.hetzner/cluster.yaml --allow-token-file
```

And without the last flag:

```
Some information in the configuration file requires your attention:
  - Hetzner token is set to a path, but paths are not allowed; use --allow-token-from-file flag
```

### DIFF
--- a/src/configuration/loader.cr
+++ b/src/configuration/loader.cr
@@ -124,7 +124,7 @@ class Configuration::Loader
   private def validate_allows_token_from_file
     if settings.hetzner_token.starts_with?("file://")
       if !allow_token_from_file
-        errors << "Hetzner token is set to a path but paths are not allowed, use --allow-token-from-file flag"
+        errors << "Hetzner token is set to a path but paths are not allowed; use --allow-token-from-file flag"
         return
       end
     end
--- a/src/configuration/loader.cr
+++ b/src/configuration/loader.cr
@@ -32,7 +32,7 @@ class Configuration::Loader
   getter settings : Configuration::Main
 
   getter hetzner_client : Hetzner::Client do
-    Hetzner::Client.new(settings.hetzner_token)
+    Hetzner::Client.new(settings.resolved_hetzner_token)
   end
 
   getter public_ssh_key_path do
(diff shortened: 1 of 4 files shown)


fabrykowski:
This functionality would be awesome!

vitobotta:
Hi @radekg and thanks a lot for your contribution!  The fix to expand the path of the config file is fine, but for the token I prefer using the environment variable when the user doesn't want to specify it in the config file. Currently, the env variable `HCLOUD_TOKEN` is supported but doesn't work because the `hetzner_token` setting is required in the config file, because I forgot to make it optional. So I would prefer fixing that so that the user would use the env variable instead of using a file somewhere in the filesystem. Could you do that instead? Thanks

vitobotta:
@radekg I have fixed the environment variable issue in 1.0.8. If you can update your PR with the other fix it would be awesome.

radekg:
Done. Apologies for the delay. I'm hiring and setting up a new office. Busy time :)

radekg:
Hi @vitobotta, no, I pushed my changes with merged upstream yesterday, but a new conflict popped up today. I merged with upstream again (https://github.com/vitobotta/hetzner-k3s/pull/153/commits/e86de6b21ddf53d464861f33a42338608a9f1da7), and all code is pushed. I was able to validate that it builds.

vitobotta:
Hi, do you have a chance to update this PR? Otherwise I will replicate the remaining change. Thanks

radekg:
Hey @vitobotta, third time lucky, maybe? I updated the PR to latest.

vitobotta:
Hi @radekg, I'm sorry for the long delay. Between some time off due to sickness and being super busy with the day job I haven't had a chance to review the remaining PR before today. Sorry about that.

Like I mentioned in a previous comment I'd like to keep the fix for relative paths, but for the token I prefer using the environment variable, which now works. I think it's more flexible and in line with 12 factor deployment ideas. Would you mind updating the PR to keep the fix for relative paths only? Thanks!

PR #186, opened 2023-03-25
Add missing reference to secret in tutorial

creib:


vitobotta:
Hi @creib , thanks for the fix!

PR #203, opened 2023-04-15
Added arm fix with pinned version

n3rdc4ptn:
I discovered a bug with the new arm servers, as the csi-driver docker image tagged with 'latest' is not compatible with arm. https://hub.docker.com/r/hetznercloud/hcloud-csi-driver/tags?page=1

I pinned the version to 2.3.2 which is compatible with arm.

Have tested it in my cluster and now the image gets pulled.


vitobotta:
Thanks for your contribution!

Issue #207, opened 2023-04-25
New ARM worker node images not matching architecture (autoscaler)


carlesjavierre:
Hello and thanks for actively maintaining this amazing project!

Here's a few log lines of the issue. It's on the cluster-autoscaler.

![image](https://user-images.githubusercontent.com/12613547/234369558-66e4955a-971b-489d-b3c8-c3a04e6f6a20.png)

The master node is created correctly, but no workers are created. It works ONLY when I create them from your script.

I've inspected the HCLOUD API and it seems like whenever you search for ubuntu, the ARM image is on the second page. Maybe since the one creating it is the cluster autoscaler it doesn't know the architecture it goes for the first one it finds? Maybe there's an architecture parameter that can be defined on the cluster-autoscaler?

This is the configuration I have on my cluster, if that helps you replicate the issue:

```yaml
hetzner_token: token
cluster_name: cluster
kubeconfig_path: "./kubeconfig"
k3s_version: v1.26.2+k3s1
public_ssh_key_path: "~/.ssh/id_ed25519.pub"
private_ssh_key_path: "~/.ssh/id_ed25519"
use_ssh_agent: false
ssh_allowed_networks:
  - 0.0.0.0/0
api_allowed_networks:
  - 0.0.0.0/0
private_network_subnet: 10.0.0.0/16
schedule_workloads_on_masters: false
image: ubuntu-22.04
masters_pool:
  instance_type: cax11
  instance_count: 1
  location: fsn1
worker_node_pools:
- name: small-autoscaled
  instance_type: cax21
  instance_count: 1
  location: fsn1
  autoscaling:
    enabled: true
    min_instances: 1
    max_instances: 5
```

vitobotta:
Ok I was able to reproduce the problem. The fix is to specify in the settings the image by ID instead of name, so for Ubuntu 22.04 on ARM it's `103908130`.  If you set the image to that value then the autoscaler correctly creates the nodes.

You can get the list of images with their IDs with the command:

```bash
curl \
	-H "Authorization: Bearer $API_TOKEN" \
	'https://api.hetzner.cloud/v1/images'
```

The problem is that until now we only had x86 nodes so my tool currently allows you to specify the image at root level in the configuration, and that setting applies to **all** nodes, included autoscaled ones. If you are planning to have a cluster with only ARM nodes this is fine, but right now you cannot have a mixed cluster because you can only specify one image.

[...]

PR #259, opened 2023-08-01
Bump cluster-autoscaler to v1.27.3

Funzinator:
The arm64 machines on Hetzner require cluster-autoscaler v1.27.1 onwards. Prior to that, the wrong architecture is sent to the API and the request fails.


flemzord:
#207

vitobotta:
Thanks for your contribution!

PR #301, opened 2023-12-14
Add NoSchedule toleration to worker upgrades (fixes #300)

### LINKED ISSUE #300: System Upgrade Controller does not target tainted nodes
When adding node taints `system-upgrade-controller` fails to upgrade the tainted nodes because there's a toleration missing for those taints.

For example:
```
- name: db
  instance_type: ccx43
  instance_count: 3
  location: ash
  labels:
  - key: node-role
    value: database
  taints:
  - key: node-role/dedicated
    value: database:NoSchedule
```

This node pool failed to upgrade. After manually adding the toleration to the created jobs the upgrade finished successfully.

Since Rancher's `system-upgrade-controller` has support for node taints in its config, I'm assuming that this is something `hetzner-k3s` would have to set based on the taints it finds in its own pools config.

lloesche:
This adds a `NoSchedule` toleration to the worker upgrades template, so that the upgrade jobs get scheduled on nodes with taints.

I have not been able to test this on an actual cluster. I left key and value as an empty string based on the following struct comments https://github.com/kubernetes/api/blob/1213fb87f503e76a8d0fde086c6d323a2cfdc644/core/v1/types.go#L3473-L3477
but don't know if maybe `system-upgrade-controller` expects the keys to be omitted entirely.


vitobotta:
Do you have a chance to test with a temp cluster?

lloesche:
Yes, I just learned how to crystal build on my local system... will create a binary of my branch and test against a temp environment when I'm at the office tomorrow. I'll ping you here when I've done that. Thank you for building hetzner-k3s btw. it's a great tool!

vitobotta:
> Yes, I just learned how to crystal build on my local system... will create a binary of my branch and test against a temp environment when I'm at the office tomorrow. I'll ping you here when I've done that. Thank you for building hetzner-k3s btw. it's a great tool!

Thank you for helping with a contribution! :)

lloesche:
Ah I just found an API key for an unused Hetzner cloud project and was able to test my change directly.

Seems to work fine. All jobs completed and the tainted nodes were upgraded from `v1.27.1-k3s1` to `v1.27.2-k3s1`.

The apply job had the following list of tolerations
```yaml
  tolerations:
  - effect: NoSchedule
    key: node.kubernetes.io/unschedulable
    operator: Exists
  - effect: NoSchedule
    operator: Exists
  - effect: NoExecute
    key: node.kubernetes.io/not-ready
    operator: Exists
    tolerationSeconds: 300
  - effect: NoExecute
    key: node.kubernetes.io/unreachable
    operator: Exists
    tolerationSeconds: 300
```

The second one didn't exist before. From my point of view the change works and solves the issue.

vitobotta:
Nice, so you didn't need to specify key and value for the taint at all? If you have tested it already and it works I can merge it already.

vitobotta:
Cool, I am merging it then. I will make a new release when I finish a couple of other things. Thanks a lot!

lloesche:
@vitobotta a new release would be much appreciated. Just ran into the same error and realized that there's not been a release since this fix was merged.

PR #304, opened 2023-12-25
Why point load balancer to nodes without the nginx ingress controller

easyStartup-pulse:
Add label to node readnme


easyStartup-pulse:
found better node selector https://pkg.go.dev/github.com/hetznercloud/hcloud-cloud-controller-manager/internal/annotation#Name
load-balancer.hetzner.cloud/node-selector

Issue #311, opened 2024-01-31
Deleting master1 destroys cluster


cwilhelm:
I have a setup of 3 master nodes and 1 worker and have a working cluster with name _franz_

```
masters_pool:
  location: fsn1
  instance_type: cx21
  instance_count: 3
worker_node_pools:
- name: test
  location: hel1
  instance_type: cx31
  instance_count: 1
```

If a delete master1 on hetzner console, and also execute `kubectl delete node _franz_-cx21-master`. I would expect to keep the same cluster with a replacement of master1 after run of `hetzner-k3s create --config ./hetzner-k3s_cluster_config.yaml`.

The script hangs in `Deploying k3s to worker franz-cx31-pool-test-worker1...` step, if i execute `kubectl get nodes` I get 
```
NAME                 STATUS   ROLES                       AGE   VERSION
franz-cx21-master1   Ready    control-plane,etcd,master   11m   v1.26.12+k3s1
```
The original cluster is gone :(

```
Validating configuration......configuration seems valid.

=== Creating infrastructure resources ===
Network already exists, skipping.
Updating firewall...done.
SSH key already exists, skipping.
Placement group franz-masters already exists, skipping.
Placement group franz-test-1 already exists, skipping.
Server franz-cx21-master2 already exists, skipping.
Server franz-cx31-pool-test-worker1 already exists, skipping.
Server franz-cx21-master3 already exists, skipping.
Creating server franz-cx21-master1...
...server franz-cx21-master1 created.
Server franz-cx21-master1 already exists, skipping.
Waiting for successful ssh connectivity with server franz-cx21-master1...
Server franz-cx21-master2 already exists, skipping.
Waiting for successful ssh connectivity with server franz-cx21-master2...
Server franz-cx21-master3 already exists, skipping.
Waiting for successful ssh connectivity with server franz-cx21-master3...
Server franz-cx31-pool-test-worker1 already exists, skipping.
Waiting for successful ssh connectivity with server franz-cx31-pool-test-worker1...
...server franz-cx21-master2 is now up.
...server franz-cx21-master3 is now up.
...server franz-cx31-pool-test-worker1 is now up.
...server franz-cx21-master1 is now up.
Load balancer for API server already exists, skipping.

=== Setting up Kubernetes ===
Deploying k3s to first master franz-cx21-master1...
[INFO]  Using v1.26.12+k3s1 as release
[...]

PR #312, opened 2024-01-31
Call first_master earlier and exclude from other_masters

cwilhelm:
get first_master earlier and exclude it from other_masters #311

### DIFF
--- a/src/cluster/create.cr
+++ b/src/cluster/create.cr
@@ -37,12 +37,12 @@ class Cluster::Create
     settings.worker_node_pools.select(&.autoscaling_enabled)
   end
   private getter kubernetes_installer : Kubernetes::Installer do
-    Kubernetes::Installer.new(configuration, masters, workers, load_balancer, ssh, autoscaling_worker_node_pools)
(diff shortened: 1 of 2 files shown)


vitobotta:
Do you mind describing these changes a bit more? How do these solve the problem you highlighted in https://github.com/vitobotta/hetzner-k3s/issues/311 exactly? 

I am not keen with introducing a "first master index" config setting" as it's brittle.

I think we can implement a more automatic recovery from this kind of scenario (first master deleted and replaced with a new one): when setting up a master which doesn't have the API responding yet, we can simply query the other masters and pick the token from the first one that responds to API calls.

So 

- new cluster, no masters active: when setting up the first master no other masters will respond to API calls yet, so assume it's a new cluster and pick the first master as main
- new cluster, first master already active while we are setting up the other masters: we repeat the above for each master identically. In this case, since the first master is assumed to be active already when we set up master2 and master3, it will pick the token from master1
- existing cluster, first master replaced: again before setting up the replacement master, we query if other masters are responding to the API, if yes we pick the token from one of them and use that token to have the replacement first master join the others instead of initializing a new cluster.

WDYT?

vitobotta:
Thanks again for the PR. I am closing because I handled this in a more robust way in https://github.com/vitobotta/hetzner-k3s/pull/348 for the upcoming new major version v2.0.0.

PR #315, opened 2024-02-02
Fix private network test ip

cwilhelm:


cwilhelm:
The fix with the 0 address improves the matter. However, the network address should be calculated correctly depending on the netmask. I'll leave that to the network professionals :)

vitobotta:
Thanks for the PR!

Issue #351, opened 2024-04-28
When changing the service_cidr, the cluster_dns address should also be changed


jpetazzo:
`service_cidr` and `cluster_dns` have default values (`10.43.0.0/16` and `10.43.0.10` respectively).

If they are both absent from the configuration, the default values are used, and the resulting cluster will have a functioning `kube-dns` Service. Yay!

However, if I set `service_cidr` (for instance, to `10.10.0.0/16`) without setting `cluster_dns`, then it looks like `cluster_dns` remains to its default value of `10.43.0.10`, and since the default value is *outside* of the service CIDR, the creation of the `kube-dns` Service fails, and the resulting cluster has a broken DNS configuration (the pods are configured with 10.43.0.10 but there is no ClusterIP service at that address).

I'm going to submit a PR to add a warning in the sample configuration file, to remind users that if they change the `service_cidr`, they should also change the `cluster_dns`.

Do you think it would be necessary to add a check in the configuration validation as well?
Or, alternatively, if `cluster_dns` is not set, should it be automatically computed from `service_cidr`?

vitobotta:
A PR would be nice. I think it's better to keep settings explicit IMO. BTW I am in the middle of a massive refactoring, and a lot of code has changed already for v2.0 which I am going to release soon. So you may want to wait for the PR. Otherwise you will need to adapt it to the new code later.

PR #352, opened 2024-04-28
Add a warning about cluster_dns and service_cidr in examples

jpetazzo:
As mentioned in #351.

There is already a note about that in the README, but I thought it might help to have it proeminently in the sample configuration as well.


vitobotta:
Ah you did it already and it's just text, so I can merge now :) Thanks!

PR #355, opened 2024-04-28
Add build instructions

jpetazzo:
I suppose that these are obvious for someone who is familiar with Crystal; but I wasn't, and it took me a while to figure out how to build the code to test my changes :)

I've also added a working_dir to the Compose file, so that the user is placed directly in the source directory. I hope this won't affect other uses (e.g. inside VSCode).

PR #357, opened 2024-04-28
Add instructions about VSCode

jpetazzo:
As requested on #355, I've added instructions about VSCode.

I had some issues running the Dev Containers extension; apparently it's not available with the open source build of VSCode; I had to install the "upstream" build and then it worked fine.


vitobotta:
Looks good, thanks!

PR #359, opened 2024-05-06
Reorder the instructions for clarity.

khustochka:
Hi, I was following the instruction how to create a basic cluster and deploy a `hello-world` application, and I encountered the same issue described in this [thread](https://github.com/vitobotta/hetzner-k3s/discussions/293): getting 400 Bad request when making http request.

I think the source of confusion here is that the step in which we check that  request to`http://hello-world.X.X.X.X.nip.io/` actually works, won't work without applying the `ConfigMap`, described in steps 19-21. I took the liberty to reorder the steps for clarity. 

Thanks!


vitobotta:
Looks good, thanks!

Issue #366, opened 2024-05-21
Failed to create Hetzner Cloud secret


vdveldet:
During deployment the following issue occurred.

```
root:/local# hetzner-k3s-linux-amd64 create --config cluster_config.yaml
Validating configuration......configuration seems valid.

=== Creating infrastructure resources ===
Network already exists, skipping.
Updating firewall...done.
SSH key already exists, skipping.
Placement group test-masters already exists, skipping.
Placement group test-small-1 already exists, skipping.
Server test-cx21-pool-small-worker1 already exists, skipping.
Server test-cx21-master1 already exists, skipping.
Server test-cx21-pool-small-worker2 already exists, skipping.
Server test-cx21-master1 already exists, skipping.
Waiting for successful ssh connectivity with server test-cx21-master1...
Server test-cx21-pool-small-worker1 already exists, skipping.
Waiting for successful ssh connectivity with server test-cx21-pool-small-worker1...
Server test-cx21-pool-small-worker2 already exists, skipping.
Waiting for successful ssh connectivity with server test-cx21-pool-small-worker2...
...server test-cx21-master1 is now up.
...server test-cx21-pool-small-worker1 is now up.
...server test-cx21-pool-small-worker2 is now up.

=== Setting up Kubernetes ===
Deploying k3s to first master test-cx21-master1...
[INFO]  Using v1.28.9+k3s1 as release
[INFO]  Downloading hash https://github.com/k3s-io/k3s/releases/download/v1.28.9+k3s1/sha256sum-amd64.txt
[INFO]  Skipping binary downloaded, installed k3s matches hash
[INFO]  Skipping installation of SELinux RPM
[INFO]  Skipping /usr/local/bin/kubectl symlink to k3s, already exists
[INFO]  Skipping /usr/local/bin/crictl symlink to k3s, already exists
[INFO]  Skipping /usr/local/bin/ctr symlink to k3s, already exists
[INFO]  Creating killall script /usr/local/bin/k3s-killall.sh
[INFO]  Creating uninstall script /usr/local/bin/k3s-uninstall.sh
[INFO]  env: Creating environment file /etc/systemd/system/k3s.service.env
[INFO]  systemd: Creating service file /etc/systemd/system/k3s.service
[INFO]  systemd: Enabling k3s unit
[INFO]  No change detected so skipping service start
Waiting for the control plane to be ready...
Saving the kubeconfig file to /local/kubeconfig...
[...]

Issue #297, opened 2023-11-29
empty kubeconfig


dyipon:
Hi,

im trying to setup a new private networked k3s cluster, but the created kubeconfig is empty. Is there any way to debug this issue?

config below:
```
---
cluster_name: hetzner-k3s
kubeconfig_path: "kubeconfig-hetzner.yaml"
k3s_version: v1.28.3+k3s1
public_ssh_key_path: "hetzner.pub"
private_ssh_key_path: "hetzner.key"

private_network_subnet: 10.0.9.0/24
existing_network: private

enable_public_net_ipv4: false
enable_public_net_ipv6: false

use_ssh_agent: false
image: ubuntu-20.04

ssh_allowed_networks:
  - 0.0.0.0/0
api_allowed_networks:
  - 0.0.0.0/0
verify_host_key: false
location: nbg1
schedule_workloads_on_masters: false

masters_pool:
  instance_type: cx11
  instance_count: 1
  location: nbg1
worker_node_pools:
- name: intel1
  instance_type: cx21
  instance_count: 1
  location: nbg1
  post_create_commands:
    - timedatectl set-timezone Europe/Berlin
    - >
      printf "network## {config## disabled}" |
      sed 's/##/:/g' > /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg
    - >
      printf "network##\n  version## 2\n  renderer## networkd\n  ethernets##\n    ens10##\n      dhcp4## true\n      nameservers##\n        addresses## [185.12.64.1, 185.12.64.2]\n      routes##\n        - to## default\n          via## 10.0.9.2" |
      sed 's/##/:/g' > /etc/netplan/50-cloud-init.yaml
    - netplan generate
    - netplan apply
    - apt update
    - apt upgrade -y
    - apt autoremove -y
post_create_commands:
  - timedatectl set-timezone Europe/Berlin
  - >
    printf "network## {config## disabled}" |
    sed 's/##/:/g' > /etc/cloud/cloud.cfg.d/99-disable-network-config.cfg
  - >
    printf "network##\n  version## 2\n  renderer## networkd\n  ethernets##\n    enp7s0##\n      dhcp4## true\n      nameservers##\n        addresses## [185.12.64.1, 185.12.64.2]\n      routes##\n        - to## default\n          via## 10.0.9.2" |
    sed 's/##/:/g' > /etc/netplan/50-cloud-init.yaml
  - netplan generate
  - netplan apply
  - apt update
[...]

PR #380, opened 2024-07-15
k3s install only AFTER cloud init is done

### LINKED ISSUE #379: Some workarounds for priv IP clusters
Hi,

since I fiddled for quite some time, maybe I can save some time for others with this.

The `enable_public_net_ipv4: false` switch did involve a few problems here, so I try list my workarounds. 
Others report that cluster creation works (e.g. #372) - but not for me and seemingly others - I keep getting the same errors, even with the exact same config than #372 :-/ (except the name of the priv network).

## Goals
- Working priv IP only cluster, configured from a bastion host with a pub ip, running that script.
- All cluster nodes can reach the internet, incl. DNS
- Autoscaler fully works, allows SSH access from bastion host

## Problems
a) k3s did not install, empty kubeconfig on bastion node, failing with STDIN error at first kubeconfig apply time - even over repeated retries
b) missing DNS on the cluster nodes
c) autoscaler did create the nodes but could not include them into the cluster
d) ssh rejects (max retries exceeded) when trying to ssh into the autoscaled nodes
e) upgrade controller was broken

## Solutions

e) was caused by [this](https://github.com/rancher/system-upgrade-controller/issues/302) and was solved by `kubectl apply -f https://github.com/rancher/system-upgrade-controller/releases/latest/download/system-upgrade-controller.yaml` after the cluster was up.  I expect rancher to fix this soon, i.e. might work when you read this out of the box (again).

All other points addressed with the following config:

```yaml
additional_packages:
 - ifupdown
 
post_create_commands:
  - printf "started" > status
  - timedatectl set-timezone Europe/Berlin
  - ip route add default via 10.0.0.1
  - ip route add 169.254.0.0/16 via 172.31.1.1
  - mkdir -p /etc/network/interfaces.d
  - echo "auto enp7s0"                                              > /etc/network/interfaces.d/enp7s0
  - echo "iface enp7s0 inet dhcp"                                  >> /etc/network/interfaces.d/enp7s0
  - echo "    post-up ip route add default via 10.0.0.1"           >> /etc/network/interfaces.d/enp7s0
  - echo "    post-up ip route add 169.254.169.254 via 172.31.1.1" >> /etc/network/interfaces.d/enp7s0
  - rm -f                            /etc/resolv.conf
  - echo 'nameserver 185.12.64.1'  > /etc/resolv.conf
  - echo 'nameserver 185.12.64.2' >> /etc/resolv.conf
  - echo 'edns edns0 trust-ad'    >> /etc/resolv.conf
  - echo 'search .'               >> /etc/resolv.conf
  - sed -i '1i export INSTALL_K3S_SKIP_DOWNLOAD=true' /root/.bashrc
  - wget 'https://github.com/k3s-io/k3s/releases/download/v1.29.6+k3s2/k3s'
  - chmod +x k3s
  - mv k3s /usr/local/bin/k3s
  - echo 'ssh-ed25519 AAAAC3Nz....  admin@bast' >> /root/.ssh/authorized_keys
  - echo 'ssh-rsa AAAAB3Nzx1yc...  me@mylaptop' >> /root/.ssh/authorized_keys
  - echo "root:somesupersecret" | chpasswd
  - printf "done" > status
```

The bastion host has 10.0.0.2 (first host created on the private network) and has a public ipv4.  On it, `echo 1 > /proc/sys/net/ipv4/ip_forward` and `iptables -t nat -A POSTROUTING -s '10.0.0.0/16' -o eth0 -j MASQUERADE` are run and persisted.

## Discussion

- The ifupdown based network setup configures the nodes to go to the internet via the bastion host, which NATs their traffic.

  - Note the `10.0.0.1` - while the bastion host is on `10.0.0.2`  - Has to be like that!
  - My images are ubuntu and debian 11 / amd.  Both have enp7s0 iface, when in the private network. Adapt if different for you.
  - The 172.31.1.1 might also be different for you - not sure. Found it via `tracepath -n 169.254.169.254` on the bastion host, first hop.

- The DNS is brutally set to point to the Hetzner's DNS servers, away from systemd resolvd, which did not resolve for queries from within containers.

- I don't reboot for install speed - but the routes and dns mod are persistent over reboots

- There is a specific route for the hetzner API endpoint at `169.254.169.254`, which the autoscaler uses to find the instance name. Without that route, the autoscaler reports something like "is this a hetzner node?" (from the [hcloud go lib](https://github.com/hetznercloud/hcloud-go/blob/main/hcloud/metadata/client.go)), in the logs and does not include new nodes into the cluster.

- The k3s is **manually** downloaded via wget, and put into place (in the matching version). Then in bashrc, the script gets a `INSTALL_K3S_SKIP_DOWNLOAD` set to true. Reason:  I keep getting failing downloads when triggered from the script - I'm sure it's not the fault of the script, works perfect with pub IPs but due to this network config. With this it works reliably, the second time you run it, after the first STDIN failure. 

Edit: See my following comment for the reason of this.

- ssh keys and root password are set from the post installer. Because for nodes created from the autoscaler, those where not set, causing hetzner to want a root password reset and so I got immediate ssh login failures (the script sets max failed attempts to a low 2).  This keys and pw setting from the post installer resolves all issues with ssh.

---

Thanks in any case for the nice installer, way easier than terraform!

axgkl:
For k3s clusers with private IPs only, cloud init is typically required, so that the nodes can reach the internet, e.g. to download k3s.

=> Lets wait until cloud init is finished, before trying to download k3s.

Fixes #379 #366 #297 and mabye a few others. With this, hetzner-k3s runs through, w/o the need to restart.

Note: The file tested is always present on hetzner cloud instances, even w/o cloud init 
=> that presence test should not cause problems elsewhere.

### DIFF
--- a/templates/master_install_script.sh
+++ b/templates/master_install_script.sh
@@ -1,8 +1,13 @@
+while [ ! -f /var/lib/cloud/instance/boot-finished ]; do
+	echo "Awaiting cloud boot-finished..."
+	sleep 5
+done
+
 touch /etc/initialized
 
-if [[ $(< /etc/initialized) != "true" ]]; then
-  systemctl restart NetworkManager || true
-  dhclient eth1 -v || true
+if [[ $(</etc/initialized) != "true" ]]; then
+	systemctl restart NetworkManager || true
+	dhclient eth1 -v || true
 fi
 
 HOSTNAME=$(hostname -f)
@@ -11,9 +16,9 @@ PUBLIC_IP=$(hostname -I | awk '{print $1}')
 NETWORK_INTERFACE=$(ip route get {{ private_network_test_ip }} | awk -F"dev " 'NR==1{split($2,a," ");print a[1]}')
 
 if [[ "{{ disable_flannel }}" = "true" ]]; then
-  FLANNEL_SETTINGS=" --flannel-backend=none --disable-kube-proxy --disable-network-policy "
+	FLANNEL_SETTINGS=" --flannel-backend=none --disable-kube-proxy --disable-network-policy "
 else
-  FLANNEL_SETTINGS=" {{ flannel_backend }} --flannel-iface=$NETWORK_INTERFACE "
+	FLANNEL_SETTINGS=" {{ flannel_backend }} --flannel-iface=$NETWORK_INTERFACE "
 fi
 
 curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION="{{ k3s_version }}" K3S_TOKEN="{{ k3s_token }}" {{ datastore_endpoint }} INSTALL_K3S_EXEC="server \
@@ -40,4 +45,4 @@ curl -sfL https://get.k3s.io | INSTALL_K3S_VERSION="{{ k3s_version }}" K3S_TOKEN
 
 systemctl start k3s # on some OSes the service doesn't start automatically for some reason
 
-echo true > /etc/initialized
+echo true >/etc/initialized


vitobotta:
Thanks again for this PR 👍

PR #391, opened 2024-07-29
Fix SSH when using non-default port

jpetazzo:
Since Ubuntu 22.10, SSH uses socket activation. This means that the SSH server isn't running by default, and gets started automatically the first time there is a connection on port 22.

Upside: it saves 3 MB of RAM.

Downside: if you're customizing the SSH port number with a drop-in configuration file at cloud-init time, it breaks SSH. SSH then needs to be restarted (or the machine needs to be rebooted) for the port number to be picked up by the systemd generator. The net result for hetzner-k3s is that if we use a non-default SSH port, provisioning breaks.

This manifests itself by the following log message:
```
  systemd[1]: ssh.socket: Socket unit configuration has changed while
  unit has been running, no open socket file descriptor left. The
  socket unit is not functional until restarted.
```
One possible fix is to disable socket activation and revert to the default mode (start SSH at boot). This can be done by disabling ssh.socket and enabling ssh.service.

This patch does exactly that, adding the corresponding commands to the cloud-init template. This has been tested with Ubuntu 24.04 and 22.04 as well.

The following link has more details on the Ubuntu change:

https://discourse.ubuntu.com/t/sshd-now-uses-socket-based-activation-ubuntu-22-10-and-later/30189/14

### DIFF
--- a/src/hetzner/instance/create.cr
+++ b/src/hetzner/instance/create.cr
@@ -212,6 +212,8 @@ class Hetzner::Instance::Create
     [
       "hostnamectl set-hostname $(curl http://169.254.169.254/hetzner/v1/metadata/hostname)",
       "update-crypto-policies --set DEFAULT:SHA1 || true",
+      "systemctl disable --now ssh.socket || true",
+      "systemctl enable --now ssh.service || true",
       "echo \"nameserver 8.8.8.8\" > /etc/k8s-resolv.conf"
     ]
   end
--- a/.gitignore
+++ b/.gitignore
@@ -29,3 +29,7 @@ temp
 docker-compose.override.yml
 cilium
 calicoctl
+
+e2e-tests/env
+e2e-tests/sshkey*
+e2e-tests/test-*
--- a/e2e-tests/README.md
+++ b/e2e-tests/README.md
@@ -1,0 +1,72 @@
+# End-to-end test harness
+
+This directory contains a few scripts to automate testing of hetzner-k3s across different combinations of configurations.
+
+## How to use this?
+
+Copy `env.sample` to `env` and edit it to indicate your Hetzner API token. You can also change the instance location if you'd like, and the location of the hetzner-k3s binary that you would like to use for testing (convenient if you're building multiple, different versions and want to check for regressions).
+
+Then to run a single test:
+
+```bash
+./run-single-test.sh config-sshport-image.yaml IMAGE=ubuntu-22.04 SSHPORT=222
+```
+
+The first argument is a configuration file template; the rest of the command line is an optional list of variables to substitute in the template.
+
+To run all the tests:
+
+```bash
+./run-all-tests.sh
+```
+
+To view test results:
+
+```
+./list-test-results.sh
+```
+
+The output will look like this:
+```
+$ ./list-test-results.sh
+config-sshport-image.yaml  IMAGE=alma-8 SSHPORT=222           error  creating   test-c59dc574
+config-sshport-image.yaml  IMAGE=alma-8 SSHPORT=22            error  creating   test-15c94339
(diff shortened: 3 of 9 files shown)


vitobotta:
Thanks for this! I had it on my list to investigate implications with hetzner-k3s since I set up a couple of Ubuntu 24.04 servers and had to make some unusual (for me) adjustments to the config in order to be able to customize the SSH port.

Before I merge, are you positive that these commands you have added also work on older versions of Debian distros as well as different types of distros? Thanks.

jpetazzo:
I've tried with ubuntu-22.04 but that's pretty much it :-)

I'll try with debian-11 and debian-12 and maybe even run a script to try with every distro that Hetzner offers and report back.

I think it should work (worst case scenario if these commands fail they shouldn't have an impact) but we'll make sure.

vitobotta:
I don't know how many people use something other than the default OS but it would be safer to test with other distros as well, not just debian based just in case. It's awesome if you can check with those too, thanks! :)

jpetazzo:
Alright, I tested with *all* the distro images offered by Hetzner, and this patch works with all of them (I can create a cluster using SSH port 222) except alma-8, rocky-8, and centos-stream-8; but these distros don't work with hetzner-k3s in the first place anyway.

Would you like a PR for the (very crude) test harness that I wrote to try many configuration combinations? :)

src/hetzner/instance/create.cr
```diff
@@ -212,6 +212,8 @@ class Hetzner::Instance::Create
     [
       "hostnamectl set-hostname $(curl http://169.254.169.254/hetzner/v1/metadata/hostname)",
       "update-crypto-policies --set DEFAULT:SHA1 || true",
+      "systemctl disable --now ssh.socket",
+      "systemctl enable --now ssh.service",
```
vitobotta, line 216: Can you change this slightly for my so these commands fail gracefully if not supported by the OS?

```suggestion
      "systemctl disable --now ssh.socket || true",
      "systemctl enable --now ssh.service || true",
```

Do you see any negative implications with this?

vitobotta:
> Alright, I tested with _all_ the distro images offered by Hetzner, and this patch works with all of them (I can create a cluster using SSH port 222) except alma-8, rocky-8, and centos-stream-8; but these distros don't work with hetzner-k3s in the first place anyway.
> 
> Would you like a PR for the (very crude) test harness that I wrote to try many configuration combinations? :)

Weird, I hadn't noticed this other message. I think you can add those changes to this PR since they are related. Thanks

src/hetzner/instance/create.cr
jpetazzo: Good idea! I made the change.

jpetazzo:
Hi!

I've updated the PR with the changes that you had suggested + the test harness that I mentioned. I wrote a small README for the test harness; happy to expand it if needed.

Cheers!

vitobotta:
Thanks a lot for the additional changes!

vitobotta:
I am actually having problems with this change. For some reason with those lines in place creating a cluster hangs randomly, while it doesn't seem to happen without. Any suggestions?

vitobotta:
After some investigation I got better results by just changing the port in the socket activation service. See https://github.com/vitobotta/hetzner-k3s/blob/more-refactoring/templates/cloud_init.yaml

jpetazzo:
Ah sorry for not seeing your message earlier!

Weird; out of curiosity, was it hanging randomly with the default port or when changing the port number?

(I'll see if I can reproduce the issue since I feel like it'll teach me a thing or two about cloud-init 😅)

vitobotta:
It was really weird, felt like some kind of race condition or something because sometimes it was working just fine, some other times ssh connections would just stop working. I am not sure why. But those issues went away with the changes I made to just change the port number in the socket related config. I would be curious to know what was the reason for that issue but I didn't spend much time on it once I figured out that setting the port made things "more stable".

PR #392, opened 2024-07-30
Add a couple of clarifying comments in sample conf

jpetazzo:
Since the main source of documentation for the configuration file format is the sample configuration, I think it's important to put a couple extra details here; namely:

- the default distro is now ubuntu-24.04
- the configuration key networking.allowed_networks.api doesn't filter access to the Kubernetes API load balancer, but to the Kubernetes API exposed on the nodes themselves

The latter is because (if I understand correctly) Hetzner load balancers don't support firewalling anyway.


vitobotta:
Thanks!

Issue #389, opened 2024-07-26
testing v2.0.0: cloud init improvement


axgkl:
Had now around 20 installs w/o a single failure.

Began to hate the output loop in cloud init. I think it's better to do it this way, for rc3, too small for a PR, except you want one. It's for master and worker:

```bash
fn_cloud="/var/lib/cloud/instance/boot-finished"
function await_cloud_init() {
	echo "🕒 Awaiting cloud config (may take a minute...)"
	while true; do
		for _ in $(seq 1 10); do
			test -f $fn_cloud && return
			sleep 2
		done
		echo -n "."
	done
}
test -f $fn_cloud || await_cloud_init
echo "Cloud init finished: $(cat $fn_cloud)"

touch /etc/initialized
```
Remove the watch if you don't like it, looks not that ugly in the terminal, except the user uses really fancy fonts, but those tend to like such stuff anyway.

Improvements:
- We save a second or 2, sleeping only 2. 
- The waiting output is **not** printed for master 2 and 3, where it is anyway finished when you apply that script.
- The output is more condensed.
- It shows the file, where the first number is the seconds it took

comes across like this:

```
...[Instance ax-master1] Waiting for successful ssh connectivity with instance ax-master1...
[Instance ax-master1] ...instance ax-master1 is now up.
[Instance ax-master2] ...instance ax-master2 is now up.
[Firewall] Updating firewall...
[Firewall] ...firewall updated
[API Load balancer] Load balancer for API server already exists, skipping create
[Instance ax-master3] 🕒 Awaiting cloud config (may take a minute...)
[Instance ax-master3] .
[Instance ax-master3] .
[Instance ax-master3] Cloud init finished: 67.21 - Fri, 26 Jul 2024 00:40:55 +0000 - v. 24.1.3-0ubuntu3.3
[Instance ax-master3] [INFO]  Using v1.30.2+k3s2 as release
[Instance ax-master3] [INFO]  Downloading hash https://github.com/k3s-io/k3s/releases/download/v1.30.2+k3s2/sha256sum-amd64.txt
[Instance ax-master3] [INFO]  Downloading binary https://github.com/k3s-io/k3s/releases/download/v1.30.2+k3s2/k3s
[Instance ax-master3] [INFO]  Verifying binary download
[Instance ax-master3] [INFO]  Installing k3s to /usr/local/bin/k3s
[Instance ax-master3] [INFO]  Skipping installation of SELinux RPM
[...]

PR #393, opened 2024-07-30
Cloud init and Timeout from 30 to 60

axgkl:
Hi, sorry, 2 things in one PR. 

The cloud init improvements we discussed in #389.

BUT also: Today I get consistently timeout errors at k3s install time.

```
[Instance ax-master2] Waiting for successful ssh connectivity with instance ax-master2...
[Instance ax-master1] Instance ax-master1 already exists, skipping create
[Instance ax-master3] Instance status: running
[Instance ax-master3] Waiting for successful ssh connectivity with instance ax-master3...
[Instance ax-master1] Instance status: running
[Instance ax-master1] Waiting for successful ssh connectivity with instance ax-master1...
[Instance ax-master2] ...instance ax-master2 is now up.
[Instance ax-master2] ...instance ax-master2 created
[Instance ax-master2] ...instance ax-master2 is now up.
Error creating instance: timeout after 00:00:30
Instance creation for ax-master3 failed. Try rerunning the create command.
Error creating instance: timeout after 00:00:30
Instance creation for ax-master1 failed. Try rerunning the create command.
[Instance ax-master1] ...instance ax-master1 is now up.
[Instance ax-master2] ...instance ax-master2 is now up.
[Instance ax-master3] ...instance ax-master3 is now up.
[Instance ax-master3] ...instance ax-master3 is now up.
```

and it does not recover from there - would have to restart the installer, then it would run through.

The only fix which allows consistently working single shot installs was to increase those 30 to 60 - then it runs through. 

Interesting is that after waiting that long the cloud init is done, when master_install runs, i.e. I don't even see the ""🕒 Awaiting cloud config (may take a minute...)" message today, just the success message from "Cloud init finished: $(cat $fn_cloud)" 

Cheers,
Gunther

PS: I can confirm the ssh port problem @jpetazzo  reports in the other PR, works only with port 22 with ubuntu.

### DIFF
--- a/src/cluster/create.cr
+++ b/src/cluster/create.cr
@@ -272,7 +272,7 @@ class Cluster::Create
         instance = nil
         begin
           Retriable.retry(max_attempts: 3, on: Tasker::Timeout, backoff: false) do
-            Tasker.timeout(30.seconds) do
+            Tasker.timeout(60.seconds) do
               instance = instance_creator.run
             end
           end
--- a/templates/master_install_script.sh
+++ b/templates/master_install_script.sh
@@ -1,7 +1,17 @@
-while [ ! -f /var/lib/cloud/instance/boot-finished ]; do
-	echo "Awaiting cloud/instance/boot-finished..."
-	sleep 5
-done
+fn_cloud="/var/lib/cloud/instance/boot-finished"
+function await_cloud_init() {
+  echo "🕒 Awaiting cloud config (may take a minute...)"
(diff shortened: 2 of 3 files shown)


vitobotta:
Thanks for the PR! I think it's fine to increase the timeout to take into account delays with the cloud init setup.

I am aware of the SSH port issue with Ubuntu 24 and will look into it as soon as I have a chance.

PR #394, opened 2024-07-31
fix: #393 My wait function definition was buggy, causing autosc. crash

axgkl:
My erroneous brackets in the function definition are detected by the syntax checker, when applied via cloud init - and cause it to crash(!) You merged #393 yesterday and did not build a release yet, so I think not many ppl have seen it - but still:

Very sorry for that!

This is taken from cloud init output log of an autoscaler created node:

```
/var/lib/cloud/instance/scripts/runcmd: 3: update-crypto-policies: not found
/var/lib/cloud/instance/scripts/runcmd: 31: Syntax error: "(" unexpected 2024-07-31 14:05:56,542 - cc_scripts_user.py[WARNING]: Failed to run module scripts_user (scripts in /var/lib/cloud/instance/scripts) 2024-07-31 14:05:56,543 - util.py[WARNING]: Running module scripts_user (<module 'cloudinit.config.cc_scripts_user' from '/usr/lib/python3/dist-packages/cloudinit/config/cc_scripts_user.py'>) failed Cloud-init v. 24.1.3-0ubuntu3.3 finished at Wed, 31 Jul 2024 14:05:56 +0000. Datasource DataSourceHetzner.  Up 71.52 seconds root@medium-autoscaled-7f6c47d6d68125cb:~#
```

### DIFF
--- a/templates/master_install_script.sh
+++ b/templates/master_install_script.sh
@@ -1,5 +1,5 @@
 fn_cloud="/var/lib/cloud/instance/boot-finished"
-function await_cloud_init() {
+function await_cloud_init {
   echo "🕒 Awaiting cloud config (may take a minute...)"
   while true; do
     for _ in $(seq 1 10); do
--- a/templates/worker_install_script.sh
+++ b/templates/worker_install_script.sh
@@ -1,5 +1,5 @@
 fn_cloud="/var/lib/cloud/instance/boot-finished"
-function await_cloud_init() {
+function await_cloud_init {
   echo "🕒 Awaiting cloud config (may take a minute...)"
   while true; do
     for _ in $(seq 1 10); do


vitobotta:
Thanks for the fix, to be honest I didn't spot any issues when I saw the previous PR and I didn't test it. Hopefully it's all good now. Thx

axgkl:
Testing... but they have problems today https://status.hetzner.com/incident/46412b10-f68b-4a84-b437-f2c14afb5449

E.g. the api lb did never get healthy allthough the ports were up, so the download of kubeconfig always timed out:

2024-07-31_1096x521_scrot

Working currently with this workaround, that works stable:

```ruby
  private def api_server_ip_address(master_count : Int)
    #master_count > 1 ? load_balancer.not_nil!.public_ip_address.not_nil! : first_master.host_ip_address.not_nil!
    first_master.host_ip_address.not_nil!
  end
```

vitobotta:
Can you please revert to the original version that was working well?

axgkl:
Hi, 

shis version w/o the () bug works as well or bad as the original. 

The problem is "only" at autoscale time and its a logical deadlock, independent of implementation: 

**Waiting indefinitely for boot-finished WITHIN the cloud init procedure must deadlock - also in the original version.**

I see 2 ways to fix: 

a) check in the bash part if the hostname contains "autoscale"
b) omit the waiting for cloud init finished, when assembling the cloud init for autoscaled nodes.


I think the b) version is more robust and I tested this here with quite a few installs and it works well:

```ruby
  private def k3s_join_script
    start_index = worker_install_script.index("touch /etc/initialized") || 0
    script_part = worker_install_script[start_index..-1]
    "|\n    #{script_part.gsub("\n", "\n    ")}"
  end
```
Yes, depends on 'touch /etc/initialized' - would you prefer a seperator than testing for this? Or better raise hard when this line is not found, so that any test will immediatlly detect it?


 I'll send a PR for this, except if you prefer a different way(?)

Issue #395, opened 2024-07-31
testing v2.0.0: autoscale node-ip


axgkl:
Hi, 
I'm testing autoscaling on priv ip clusters.

After this https://github.com/vitobotta/hetzner-k3s/pull/394#issuecomment-2261144466 the autoscaler creates the nodes and they join the cluster. All kube-system pods for the new ones running.

Only left problem: Scheduling 6 pods with an affinity rule that they should be on 6 differnent hosts, only the 3 on my 3 masters get into running - the others remain pending.

ccm log says:

```
E0731 20:39:48.001894       1 node_controller.go:240] error syncing 'medium-autoscaled-2f625ed856f50e8': failed to get node modifiers from cloud provider: provided node ip for node "medium-autoscaled-2f625ed856f50e8" is not valid: failed to get node address from cloud provider that matches ip: 100.66.1.123, requeuing
```

And those nodes kept having the unschedulable taint.

Then  I went on an autoscaled node via ssh and reconfigured the node-ip in /etc/systemd/system/k3s-agent.service away from the 100.66.1.123 to the ip within the private network (the 10.1.0.10 below) - and restarted the service.
 
 => Instantly working pod, i.e. that was the culprint.

```
 
❯ kc get nodes -o wide                                                                               !?
NAME                                 STATUS   ROLES                       AGE     VERSION        INTERNAL-IP      EXTERNAL-IP   OS-IMAGE           KERNEL-VERSION     CONTAINER-RUNTIME
ax-master1                           Ready    control-plane,etcd,master   24m     v1.30.2+k3s2   10.1.0.5         <none>        Ubuntu 24.04 LTS   6.8.0-38-generic   containerd://1.7.17-k3s1
ax-master2                           Ready    control-plane,etcd,master   24m     v1.30.2+k3s2   10.1.0.3         <none>        Ubuntu 24.04 LTS   6.8.0-38-generic   containerd://1.7.17-k3s1
ax-master3                           Ready    control-plane,etcd,master   25m     v1.30.2+k3s2   10.1.0.4         <none>        Ubuntu 24.04 LTS   6.8.0-38-generic   containerd://1.7.17-k3s1
medium-autoscaled-16f63cb4d524e27b   Ready    <none>                      5m43s   v1.30.2+k3s2   10.1.0.10        <none>        Ubuntu 24.04 LTS   6.8.0-38-generic   containerd://1.7.17-k3s1
[...]

PR #396, opened 2024-07-31
Fixes for installer scripts

axgkl:
#### 1. Make install scripts POSIX compliant, so that they run for autoscaled nodes in priv ip situations (and also on minimal distris w/o bash). 

Note: Besides the `if [[ ..]]` checks which were failing on autoscaled nodes: What I thought was a bug in `function await_cloud_init () { ...}` is also just a bashism, like all the others in that script - it IS allowed in bash (that's why my LSP did not complain).
Removing the brackets made it working, since it *is* accepted by the minimal shell in which cloud-init runs,. But it is still not posix compliant - there it must be `await_cloud_init() { ...}` .  

**Affects**
- templates/master_install_script.sh
- templates/worker_install_script.sh

**Ref**  #395

----

#### 2. Prevent the cloud init finished detection to be part of the cloud init on **autoscaled** nodes.  

I did it by cutting off all before 'touch /etc/initialized', which is a bit risky regarding changes of that line later, so I also  added a check if that line is really in, hard failing if not, so one will see the effect immediately and can adapt.

Good news: We have **no**(!) problem with the current v1 release - there the waiting for cloud init is NOT in the worker_install_script.sh, which is merged into the cloud init of autoscaled ones. It is only in the master install script.
Of course worker creation might fail on non autoscaled workers, w/o the waiting for finished, but in v1 it always was like this.

**Affects**
- src/kubernetes/software/cluster_autoscaler.cr

**Ref**  #394

----

#### 3. Solve timeout when api_server_hostname is given.

[docs](https://github.com/vitobotta/hetzner-k3s/blob/v2.0.0.rc1/docs/Creating_a_cluster.md) say `# api_server_hostname: k8s.example.com # optional: DNS for the k8s API LoadBalancer. After the script has run, create a DNS record with the address of the API LoadBalancer.`

And indeed, the user can't do this BEFORE running the script, not knowing the IP of the API LB created.

But since we set that hostname into the kubeconfig, when we do `save_kubeconfig(master_count)` and in the next command do `kubectl cluster-info`, this can't work - since at that time the DNS is not configured to point to that api loadbalancer.

My suggested fix is to set the **IP** of the LB into the kubeconfig. Then all will work and the user can, at his pace, configure his DNS to point to the api lb for that hostname - and only then adapt the kubeconfig, if wanted. SSL will work since we configure tls-sans for that hostname anyway.
But kubeconfig, we can't do this for the user, having no control over if and when he configures his DNS. 

**Affects**
- src/kubernetes/installer.cr

**Ref** None, no issue created.

----

#### 4. Suggested Feature

Vito, would you accept a PR for the feature below, did not dare to put it in here - but it would help me a lot.

Thing is: I don't want the API loadbalancer, keep deleting it anyway and point my kubectl to one of the masters. 

So this would allow to have it set to the first master via the "api_server_hostname" key:

```diff
diff --git a/src/kubernetes/installer.cr b/src/kubernetes/installer.cr
index 8f33273..27c6735 100644
--- a/src/kubernetes/installer.cr
+++ b/src/kubernetes/installer.cr
@@ -358,6 +358,10 @@ class Kubernetes::Installer
   end

   private def api_server_ip_address(master_count : Int)
-    master_count > 1 ? load_balancer.not_nil!.public_ip_address.not_nil! : first_master.host_ip_address.not_nil!
+    if settings.api_server_hostname == "first_master"
+      first_master.host_ip_address.not_nil!
+    else
+      master_count > 1 ? load_balancer.not_nil!.public_ip_address.not_nil! : first_master.host_ip_address.not_nil!
+    end
   end
 end
 ```

 
 Add Note: Yesterday, all day, the LB status never(!) went healthy, allthough 6443 ports where all listening,
 hetzner had problems. This feature here allows installs nevertheless, while with the current version there is no way, in such a situation, to get the installer running through. Not sure naturally how often this happens. 
 
2024-07-31_1096x521_scrot

### DIFF
--- a/src/kubernetes/installer.cr
+++ b/src/kubernetes/installer.cr
@@ -276,7 +276,7 @@ class Kubernetes::Installer
     log_line "Saving the kubeconfig file to #{kubeconfig_path}...", "Control plane"
 
     kubeconfig = ssh.run(first_master, settings.networking.ssh.port, "cat /etc/rancher/k3s/k3s.yaml", settings.networking.ssh.use_agent, print_output: false).
-      gsub("127.0.0.1",  settings.api_server_hostname ? settings.api_server_hostname : api_server_ip_address(master_count)).
+      gsub("127.0.0.1",  api_server_ip_address(master_count)).
       gsub("default", settings.cluster_name)
 
     File.write(kubeconfig_path, kubeconfig)
--- a/src/kubernetes/software/cluster_autoscaler.cr
+++ b/src/kubernetes/software/cluster_autoscaler.cr
@@ -39,7 +39,13 @@ class Kubernetes::Software::ClusterAutoscaler
   end
 
   private def k3s_join_script
-    "|\n    #{worker_install_script.gsub("\n", "\n    ")}"
+    start_index = worker_install_script.index("touch /etc/initialized") || 0
+    # make sure we early detect, when this line would be changed in the worker install script.
+    # Keeping "cloud init finished"-detection within the autoscaled nodes would deadlock,
+    # since there we run it *during* cloud init.
+    raise "Error: 'touch /etc/initialized' not found in worker_install_script" unless start_index
+    script_part = worker_install_script[start_index..-1]
+    "|\n    #{script_part.gsub("\n", "\n    ")}"
   end
 
   private def certificate_path
--- a/templates/master_install_script.sh
+++ b/templates/master_install_script.sh
@@ -15,7 +15,7 @@ echo "Cloud init finished: $(cat $fn_cloud)"
 
 touch /etc/initialized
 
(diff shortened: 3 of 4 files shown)


vitobotta:
I appreciate A LOT the help you are giving to improve and fix things while we finalize the v2 version. However it would be easier if you created a separate PR for each issue otherwise it's difficult to review and test each change in isolation. Any change you could split this PR into one per issue? :)

axgkl:
Hi, ok I'll try but I'm tense the next days, will happen until the weekend.

vitobotta:
> Hi, ok I'll try but I'm tense the next days, will happen until the weekend.

Sounds good, thanks!

PR #405, opened 2024-08-07
fix: Filter cloud init wait

axgkl:
Autoscaled nodes run the worker install script *within* cloud init itself.

There we can't wait for the file demarking finished, will never happen, since we are not finished ->
we are stuck, waiting, deadlocked.

### DIFF
--- a/src/kubernetes/software/cluster_autoscaler.cr
+++ b/src/kubernetes/software/cluster_autoscaler.cr
@@ -40,7 +40,13 @@ class Kubernetes::Software::ClusterAutoscaler
   end
 
   private def k3s_join_script
-    "|\n    #{worker_install_script.gsub("\n", "\n    ")}"
+    start_index = worker_install_script.index("touch /etc/initialized") || 0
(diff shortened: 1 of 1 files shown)


axgkl:
(won't send the posix compliancy PR, seeing that you integrated it already)

vitobotta:
Just tested this and it works, thanks! I think we need to find a more robust way to handle this, like instead of having the cloud init wait part directly in the worker script, which is included in the cloud init, we run it separately as the first script, before the current worker install script, and only for static - not autoscaled - nodes, where we know that the k3s installation has to be done after cloud init is complete. Thanks again :)

axgkl:
right. that wait has to be run only for static non autoscaled nodes. i don't like that split either, that's technical debt, so totally liking the idea of a seperate waiting step, before applying the installer.

vitobotta:
I have made the change I mentioned and now the cloud init wait script is run separately. Will push it as rc5 as soon as I have finished another couple of things.

Issue #408, opened 2024-08-08
post_create_commands prevent autoscaler from working


Jboulery:
Hey,

I have an existing cluster on Hetzner and wanted to use a gateway / NAT server to have a fixed IP that I can whitelist in third party services (e.g. databases).
Followed [this Hetzner community tutorial](https://community.hetzner.com/tutorials/how-to-set-up-nat-for-cloud-networks).

Original config (autoscaling works fine):
```yaml
hetzner_token: ...
cluster_name: x-dev
kubeconfig_path: "..."
k3s_version: v1.30.3+k3s1
public_ssh_key_path: "..."
private_ssh_key_path: "..."
use_ssh_agent: false
ssh_allowed_networks:
  - 0.0.0.0/0
api_allowed_networks:
  - 0.0.0.0/0
private_network_subnet: 10.0.0.0/16
schedule_workloads_on_masters: false
masters_pool:
  instance_type: cpx21
  instance_count: 1
  location: nbg1
worker_node_pools:
  - name: autoscaled-cpx41
    instance_type: cpx41
    instance_count: 2
    location: nbg1
    autoscaling:
      enabled: true
      min_instances: 0
      max_instances: 5
```

New config (autoscaling stops working correctly):
```yaml
hetzner_token: ...
cluster_name: x-dev
kubeconfig_path: "..."
k3s_version: v1.30.3+k3s1
public_ssh_key_path: "..."
private_ssh_key_path: "..."
use_ssh_agent: false
ssh_allowed_networks:
  - 0.0.0.0/0
api_allowed_networks:
  - 0.0.0.0/0
private_network_subnet: 10.0.0.0/16
schedule_workloads_on_masters: false
additional_packages:
  - ifupdown
post_create_commands:
  - apt update
  - apt upgrade -y
  - apt autoremove -y
  - |
    cat <<'EOF' >> /etc/systemd/network/10-ens10.network
    [Match]
    Name=ens10
    
    [Network]
    DHCP=yes
    Gateway=10.0.0.1
    EOF

masters_pool:
  instance_type: cpx21
  instance_count: 1
  location: nbg1

worker_node_pools:
  - name: autoscaled-cpx41
    instance_type: cpx41
    instance_count: 2
    location: nbg1
    autoscaling:
      enabled: true
      min_instances: 0
      max_instances: 5
```

Output:
```
Validating configuration......configuration seems valid.

=== Creating infrastructure resources ===
Network already exists, skipping.
[...]

PR #417, opened 2024-08-20
Add some NAT doc in Setting up a cluster.md

Jboulery:
Some documentation to make the autoscaler route properly if a NAT should be used.

Relates to https://github.com/vitobotta/hetzner-k3s/issues/408


vitobotta:
Thanks!

PR #431, opened 2024-08-31
Nginx default ingress class

lauw70:
After many hours of staring at a 404 Not Found page from Nginx, I found there is such a thing as a "default Ingress Class".
I propose we make this part of the setup example, as the majority of the users expect this by default.
Additionally, many Helm Charts do not specify an Ingress Class, resulting in a 404 Not Found page.

(I also added a link to this setup doc, as I spent many hours today trying to get it working using snippets from issues, only to find out that there was a really nice guide hidden in the docs folder)


vitobotta:
Nice PR, thanks!

PR #437, opened 2024-09-04
Fix e2e-tests in devcontainer

ValentinVoigt:
These two packages are needed to run e2e-tests inside the devcontainer.


vitobotta:
Thanks!

PR #438, opened 2024-09-04
Fix #434 (subnet sizes != /16)

### LINKED ISSUE #434: Unable to use private networks with subnet size != 16
I have an existing network at Hetzner, to which I need to add a new cluster. I use the `existing_network_name`-feature for that, as per the docs:

```yaml
networking:
  private_network:
    enabled: true
    subnet: 172.16.0.0/12
    existing_network_name: "heo"
```

Unfortunately, since what I think was [this commit](https://github.com/vitobotta/hetzner-k3s/commit/91eed4c027b42123e0d279113fcd1fc08b119831l), I am unable to add a new cluster. The relevant excerpt from the logs is:

```
[Instance prod-cx22-master2] Waiting for private network IP in subnet 172.16.0.0/12 to be available... (Attempt 1/30)
[Instance prod-cx22-master2] Waiting for private network IP in subnet 172.16.0.0/12 to be available... (Attempt 2/30)
[Instance prod-cx22-master2] Waiting for private network IP in subnet 172.16.0.0/12 to be available... (Attempt 3/30)
^C
```

After some digging, it looks like the following is happening: the `master_install_script.sh` [here](https://github.com/vitobotta/hetzner-k3s/blob/cc73a0aaf303c1280228bbc57a22f85490a5c4de/templates/master_install_script.sh#L9) script takes my `172.16.0.0/12` subnet, removes the `/12` and removes the last `.0` making it `172.16.0.`. As my subnet is actually a `/12` and the server's new IP address is `172.18.X.Y` (which is indeed part of the original `/12`), a simple `grep` will not match. This would be true for every subnet that is not a `/16`.

The same probably applies to `worker_install_script.sh` as well, although I haven't tried.

I think there should be a hint in the docs, if this is not going to be supported anymore.

Adopting the code to just support different sizes might be very difficult. There is some awk-magic around or one could use some inline Python. But maybe there's a better way than using grep on ip. 

I don't know the code's intention to reasonably argue about that, but I may have a suggestion. Hetzner writes [here, that their private network MTU is always 1450](https://docs.hetzner.com/cloud/technical-details/faq#what-is-the-mtu-setting-for-cloud-server-network-interfaces), while the public interface has a MTU of 1500. Instead of looking for a matching IP address, maybe the following could be an alternative?

```bash
ip -4 addr show | egrep -q "mtu 1450.+state UP"
```

ValentinVoigt:
This should fix #434. I took the freedom to simplify the command you proposed a bit. I also removed `private_network_test_ip`, which had the same problem with subnet size. My IDE ran `shfmt` on the `.sh`-files; I kept the reformatting, if you don't mind.

I also tried running e2e-tests (see GH-437), but stopped the execution, when tests where spamming dozens of errors quickly (`image 'fedora-38' not found`). I also had random connection issues after server creation. I did not investigate further at this point. I don't think they were related to my changes.

Random connection problems:

```
Error creating instance: timeout after 00:01:00
Instance creation for test-e9acedda-master1 failed. Try rerunning the create command.
```

Aborted test results:

```
config-sshport-image.yaml  IMAGE=alma-8 SSHPORT=22         error    done      test-15c94339
config-sshport-image.yaml  IMAGE=alma-8 SSHPORT=222        error    done      test-c59dc574
config-sshport-image.yaml  IMAGE=alma-9 SSHPORT=22         ok       done      test-3a378dbe
config-sshport-image.yaml  IMAGE=alma-9 SSHPORT=222        error    done      test-e9acedda
config-sshport-image.yaml  IMAGE=debian-11 SSHPORT=22      ok       done      test-77bf45fe
config-sshport-image.yaml  IMAGE=debian-11 SSHPORT=222     ok       done      test-fe655f1c
config-sshport-image.yaml  IMAGE=debian-12 SSHPORT=222     ok       done      test-e16755b9
config-sshport-image.yaml  IMAGE=fedora-38 SSHPORT=222     pending  creating  test-06bdf688
config-sshport-image.yaml  IMAGE=rocky-8 SSHPORT=22        ok       done      test-34ca1517
config-sshport-image.yaml  IMAGE=rocky-8 SSHPORT=222       error    done      test-48a902f8
config-sshport-image.yaml  IMAGE=rocky-9 SSHPORT=22        ok       done      test-3c4c5d45
config-sshport-image.yaml  IMAGE=ubuntu-20.04 SSHPORT=222  ok       done      test-9b2fbf59
config-sshport-image.yaml  IMAGE=ubuntu-22.04 SSHPORT=222  ok       done      test-86176827
config-sshport-image.yaml  IMAGE=ubuntu-24.04 SSHPORT=222  ok       done      test-b7c132d6
```


vitobotta:
Thanks for the PR! It's a bit delicate change so I will need to test it throughly when I have more time.

ValentinVoigt:
I very much agree with you. This should not be rushed.

vitobotta:
Hi @ValentinVoigt , sorry for the long delay. Work has been keeping me busy, so I haven’t had much time to focus on the project until this week. Now that I’m back, I’ve been working on a lot of things, but among other things I had to do a force push to remove a file that accidentally got published. 

PR #458, opened 2024-10-02
Adds custom timeout for instance creation

fjaeger-seibert:
If I try to create a cluster in a private network only, the cloud init takes a very long time. Up to 6 minutes, so it would be helpful to be able to adjust the timeout.

I hope my changes are ok, never worked with Crystal before :D

### DIFF
--- a/README.md
+++ b/README.md
@@ -45,8 +45,6 @@ Using `hetzner-k3s`, creating a highly available k3s cluster with 3 masters for 
 
 Also see this [documentation page](https://github.com/vitobotta/hetzner-k3s/blob/main/docs/Setting%20up%20a%20cluster.md) for a tutorial on how to set up a cluster with the most common setup to get you started.
 
-If you like this project and would like to help its development, consider [becoming a sponsor](https://github.com/sponsors/vitobotta).
-
 ___
 ## Who am I?
 
@@ -81,6 +79,7 @@ A huge thank you to [**Alamos GmbH**](https://alamos.gmbh) for sponsoring this p
 
 Also thanks to [@magicpages](https://github.com/magicpages), [@jonasbadstuebner](https://github.com/jonasbadstuebner), [@worldsayshi](https://github.com/worldsayshi), [@QuentinFAIDIDE](https://github.com/QuentinFAIDIDE) and the other sponsors. Your contributions are very much appreciated!
 
+If you like this project and would like to help its development, consider [becoming a sponsor](https://github.com/sponsors/vitobotta) too.
 ___
 ## Code of conduct
 


vitobotta:
Hi! Thanks for the PR. It looks OK and it doesn't seem to be risky to merge, but I will test a bit as soon as I have a chance before making a release. Would you mind describing your setup and configuration as detailed as possible so I can reproduce the context easily? Thanks!

vitobotta:
Also, did you figure out why the cloud init takes a long time in this case?

vitobotta:
Can you also describe more the s2s router setup as well? Just trying to think what might be the problem.

fjaeger-seibert:
The s2s router is the connection to the rest of the VPN network. I can use it to reach VMs without public IPs from the user VPN. The monitoring and trending of the systems also runs via it
It also makes the router for internal IPs to the outside. It runs a Wireguard client and otherwise only Linux routing

KristijanL:
I am facing the same issues with K3s on Hetzner with a private network only. The reason it takes so long is that the default Hetzner images are configured with DHCP enabled by default. However, the “post_create_commands” are only executed after the VM boots up with the default network configuration. During boot, the system waits until network-related timeouts occur, which causes delays. Once the VM successfully boots, the network is reconfigured and starts working properly. My workaround is to prepare a snapshot with the network already configured.

vitobotta:
> I am facing the same issues with K3s on Hetzner with a private network only. The reason it takes so long is that the default Hetzner images are configured with only DHCP enabled and no custom default route / dns by default. However, the “post_create_commands” are only executed after the VM boots up with the default network configuration. During boot, the system waits until network-related timeouts occur, which causes delays. Once the VM successfully boots, the network is reconfigured and starts working properly. My workaround is to prepare a snapshot with the network already configured.

That makes a lot of sense, thanks for looking into it! I could add a new option to the config settings `pre_k3s_setup_commands` or something similar, to allow for configuring commands to be run before setting up k3s, such as networking related ones. Would this suffice or do you have better suggestions?

KristijanL:
I would propose modifying the function as follows:
`combined_commands = [formatted_additional_commands, post_create_commands, init_commands].flatten`

This change ensures that the execution flow remains consistent with the master init. With this modification, all post_create_commands are executed before the SSH script is run on the master init node as well as on worker nodes.

KristijanL:
After further testing, I discovered that all Ubuntu images perform a network test, which causes them to wait for a timeout. In contrast, Debian 11 and Debian 12 do not have this issue and are deployed within the expected timeframes.

The preconfigured images, however, have a significant flaw: their deployment times are inconsistent, from under 1 minute to 5 minutes or more. This unpredictability makes them unsuitable for scenarios requiring rapid resource scaling.

vitobotta:
Hi @fjaeger-seibert, sorry for the long delay. Work has been keeping me busy, so I haven’t had much time to focus on the project until this week. Now that I’m back, I’ve been working on a lot of things, but among other things I had to do a force push to remove a file that accidentally got published. 

PR #471, opened 2024-10-21
add support for etcd backups with s3 storage

Kennybll:
K3S has built in support for uploading snapshots of the embedded etcd database to S3. This PR adds the settings to enable that to the configuration file. 

If the s3 settings are not included it will default to not being enabled, to ensure backward compatibility. 

I also added a check to make sure it is only enabled when using the etcd datastore mode.

It may also work with other S3 compatible providers. I cannot give a list, however I tested it with Backblaze B2 and it worked.

An example when uploading to AWS S3 in the us-east-1 region:
```yaml
datastore:
  mode: etcd
  s3:
    enabled: true
    access_key: EXAMPLE_ACCESS_KEY
    secret_key: EXAMPLE_SECRET_KEY
    bucket: EXAMPLE_BUCKET
```


src/configuration/datastore.cr
```diff
@@ -1,11 +1,14 @@
 require "yaml"
+require "./s3"
 
 class Configuration::Datastore
   include YAML::Serializable
 
   getter mode : String = "etcd"
   getter external_datastore_endpoint : String = ""
 
+  getter s3 : Configuration::S3 = Configuration::S3.new
```
vitobotta, line 10: By convention, I'd add an intermediate namespace as in `Configuration::Datastore::S3` since this configuration is specific to the datastore only.

src/configuration/settings/datastore.cr
```diff
@@ -8,8 +10,10 @@ class Configuration::Settings::Datastore
   def validate
     case datastore.mode
     when "etcd"
+      Settings::S3.new(errors, datastore.s3).validate
```
vitobotta, line 13: This also should ideally be `Settings::Datastore::S3`

src/configuration/settings/s3.cr
```diff
@@ -0,0 +1,16 @@
+class Configuration::Settings::S3
+  getter errors : Array(String)
+  getter s3 : Configuration::S3
+
+  def initialize(@errors, @s3)
+  end
+
+  def validate
+    case s3.enabled
```
vitobotta, line 9: Perhaps you could simplify this a bit since you are only checking one condition and the `case` statement is overkill.

```suggestion
    return unless s3.enabled
    
    errors << "access_key is required for S3 backups" if s3.access_key.strip.empty?
    errors << "secret_key is required for S3 backups" if s3.secret_key.strip.empty?
    errors << "bucket is required for S3 backups" if s3.bucket.strip.empty?
```

src/kubernetes/installer.cr
```diff
@@ -183,6 +183,7 @@ class Kubernetes::Installer
       cluster_dns: settings.networking.cluster_dns,
       datastore_endpoint: datastore_endpoint,
       etcd_arguments: etcd_arguments,
+      s3_arguments: generate_s3_arguments(),
```
vitobotta, line 186: Similarly to Ruby, there is no need for `()` when invoking a method without any arguments :)

vitobotta: Also I'd rename this to `etcd_backup_settings` or `etcd_snapshot_settings` to make it clear that these settings are for etcd backups/snapshots.

vitobotta (changes requested):
Wow, you made me VERY happy with this PR! It was on my list, so I appreciate you took the time to implement it. 

I requested some changes if you don't mind, but these are mostly minor things and quick changes as you have done a great job really.

Thanks a lot!

docs/Creating_a_cluster.md
```diff
@@ -47,6 +47,18 @@ networking:
...
+  #   endpoint: "s3.amazonaws.com" # optional defaults to s3.amazonaws.com
+  #   endpoint_ca: "" # optional
+  #   skip_ssl_verify: false # optional defaults to false
+  #   access_key: ""
+  #   secret_key: ""
+  #   bucket: ""
+  #   region: "us-east-1" # optional defaults to us-east-1
+  #   folder: "" # optional
+  #   insecure: false # optional defaults to false
+  #   timeout: "5m0s" # optional defaults to 5m0s
```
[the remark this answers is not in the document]
Kennybll: I agree about grouping the settings, I decided to put the format below. That way it allows for extending the different options in the future. I also added some of the backup specific options, not just s3. So retention, what folder to store the backups, etc.

```yaml
datastore:
  etcd:
    backups:
      s3:
        enabled: true
        ......: ...
```

Kennybll:
I realize it was outside the scope of this PR, but after requesting to refactor the structor of the cluster_config.yaml options I figured I would add some of the additional back up options.

The only backup option I could not figure out was schedule_cron. I could not get it to work correctly for the life of me. Maybe it's the way the string was getting parsed with all of the quotes. That can be saved for a future PR. Note to self, I was looking at creating a /etc/rancher/k3s/config.yaml file with the snapshot-cron value, instead of passing it as an arg to the k3s binary.

src/kubernetes/installer.cr
```diff
@@ -388,4 +389,33 @@ class Kubernetes::Installer
       first_master.private_ip_address
     end
   end
+
+  private def etcd_backup_settings
+    if !settings.datastore.etcd.backups.enabled
```
vitobotta, line 394: You can also do

```suggestion
    unless settings.datastore.etcd.backups.enabled
```

src/kubernetes/installer.cr
```diff
@@ -388,4 +389,33 @@ class Kubernetes::Installer
...
+      return "--etcd-disable-snapshots"
+    end
+
+    opts = [] of String
+
+    backups = settings.datastore.etcd.backups
+    opts << "--etcd-snapshot-retention=#{backups.retention}" if backups.retention != nil
```
vitobotta, line 401: Why not use the `present?` helper here too?

docs/etcd S3 backups.md
```diff
@@ -0,0 +1,5 @@
+### etcd S3 Backups
+
+S3 backups can be enabled for the embedded etcd mode only. You can see the explainations for each option in the [K3s docs](https://docs.k3s.io/cli/etcd-snapshot).
+
+> "In addition to backing up the datastore itself, you must also back up the server token file at /var/lib/rancher/k3s/server/token. You must restore this file, or pass its value into the --token option, when restoring from backup. If you do not use the same token value when restoring, the snapshot will be unusable, as the token is used to encrypt confidential data within the datastore itself." [K3S Backup/Restore Docs](https://docs.k3s.io/datastore/backup-restore)
```
vitobotta, line 5: Since you have been working on this awesome PR already, would you be up to adding logic to also extract the token file automatically like we do for the kubeconfig? If you can't I will look into it when I test the PR :)

vitobotta (changes requested):
A few more comments, it's looking good!!! 👍

docs/Creating_a_cluster.md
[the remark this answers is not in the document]
vitobotta: This is perfect, thanks!

src/configuration/datastore.cr
vitobotta: The hierarchy of the classes looks much better now 👍

README.md
```diff
@@ -57,6 +57,8 @@ See my public profile with links for connecting with me [here](https://vitobotta
 - [Load balancers](docs/Load_balancers.md)
 - [Storage](docs/Storage.md)
 - [Troubleshooting](docs/Troubleshooting.md)
+- [S3 Backups](docs/S3%20backups.md)
```
vitobotta, line 60: You left the old link :)

src/configuration/settings/datastore.cr
vitobotta: Did you accidentally remove the `validate` method? I don't see it here anymore

vitobotta: Nevermind I was looking at the wrong file.

vitobotta:
> I realize it was outside the scope of this PR, but after the request to refactor the structor of the cluster_config.yaml options I figured I would add some of the additional back up options.
> 
> The only backup option I could not figure out was schedule_cron. I could not get it to work correctly for the life of me. Maybe it's the way the string was getting parsed with all of the quotes. That can be saved for a future PR. Note to self, I was looking at creating a /etc/rancher/k3s/config.yaml file with the snapshot-cron value, instead of passing it as an arg to the k3s binary.

Can you share more details on the issue you had with that setting? It should be a regular string if it's in cron format, isn't that the case?

src/kubernetes/installer.cr
Kennybll: I didn't use the `present?` helper since it takes in a `String | Nil`, but retention is a `Int32`

docs/etcd S3 backups.md
Kennybll: Yeah, I can attempt that

src/kubernetes/installer.cr
vitobotta: Gotcha!

docs/etcd S3 backups.md
vitobotta: That would be awesome, thanks!

Kennybll:
> > The only backup option I could not figure out was schedule_cron. I could not get it to work correctly for the life of me. Maybe it's the way the string was getting parsed with all of the quotes. That can be saved for a future PR. Note to self, I was looking at creating a /etc/rancher/k3s/config.yaml file with the snapshot-cron value, instead of passing it as an arg to the k3s binary.
> 
> Can you share more details on the issue you had with that setting? It should be a regular string if it's in cron format, isn't that the case?

I tried every variation I could think of for a cli flag. None of them worked. I even tried manually setting it in master_install_script.sh in case it was the YAML parser being weird. If you are able to get it to work that would be awesome as it would be cleaner.
`--etcd-snapshot-schedule-cron="* * * * *"`
`--etcd-snapshot-schedule-cron='* * * * *'`
`--etcd-snapshot-schedule-cron=* * * * *`
`--etcd-snapshot-schedule-cron * * * * *`

Based on this comment https://github.com/k3s-io/k3s/issues/5983#issuecomment-1215741365, I was able to get it to work using a config file. So if `snapshot-cron` is set, the master_install_script.sh will create a file `/etc/rancher/k3s/config.yaml.d/hetzner-k3s.yaml` with the flag. If the option is removed, then it will delete that config file if it exists.

Whenever a cli flag changes the k3s service will automatically restart when rerunning the master_install_script.sh. However, if a config file changes, the service will not restart. I created an issue in the k3s repo https://github.com/k3s-io/k3s/issues/11180. But in the meantime we would have to manually restart the k3s service, just in case there are changes in a config file, even if there are none.

Kennybll:
Yes, I tried escaping with \" and also \\\" just in case

vitobotta:
> Yes, I tried escaping with \" and also \\\" just in case

I see. Thanks for the clarification

Thijmen:
I saw this PR and I am very curious about this. What is the timeline for this PR, when can we expect it in a possible release?

vitobotta:
> I saw this PR and I am very curious about this. What is the timeline for this PR, when can we expect it in a possible release?

I can't say exactly when it'll be done because I have some other important things to take care of right now, but it will definitely be part of one of the upcoming releases.

Thijmen:
@vitobotta thanks for getting back with me, I'll just keep an eye on this PR in that case! :)

vitobotta:
> @vitobotta thanks for getting back with me, I'll just keep an eye on this PR in that case! :)

Np :)

psavva:
I would also like to see this merged :)

vitobotta:
Hi @Kennybll, sorry for the long delay. Work has been keeping me busy, so I haven’t had much time to focus on the project until this week. Now that I’m back, I’ve been working on a lot of things, but among other things I had to do a force push to remove a file that accidentally got published. 

vitobotta:
Done in v2.3.6.

PR #480, opened 2024-11-07
Switch to standard cluster-autoscaler image

ghaering:
Hetzner started sending emails to all customers using the docker.io/hetznercloud/cluster-autoscaler images. The email reads that these images were only provided temporarily until offical releases by upstream cluster-autoscaler were ready. They ask us to start using the official images and that they will REMOVE their own images on November 19 2024.

I have tested this by rebuilding the CLI tool, creating a cluster and creating a deployment that triggered the autoscaler.

### DIFF
--- a/src/kubernetes/software/cluster_autoscaler.cr
+++ b/src/kubernetes/software/cluster_autoscaler.cr
@@ -94,7 +94,7 @@ class Kubernetes::Software::ClusterAutoscaler
   end
 
   private def patch_autoscaler_container(autoscaler_container)
-    autoscaler_container.image = "docker.io/hetznercloud/cluster-autoscaler:v1.31.0-hcloud1"
+    autoscaler_container.image = "registry.k8s.io/autoscaling/cluster-autoscaler:v1.31.1"
     autoscaler_container.command = container_command
 
     set_container_environment_variable(autoscaler_container, "HCLOUD_CLOUD_INIT", Base64.strict_encode(cloud_init))


trinhthinh388:
I wonder if this will work with the already created cluster as well?

ghaering:
@trinhthinh388 I don't know and have not tried yet. But the autoscaler in your existing clusters will stop working in two weeks if it happens to be rescheduled to a different node.

As a workaround, you go to your existing cluster and "kubectl edit deployment" the cluster autoscaler and replace the image. This works fine.

trinhthinh388:
> @trinhthinh388 I don't know and have not tried yet. But the autoscaler in your existing clusters will stop working in two weeks if it happens to be rescheduled to a different node.
> 
> As a workaround, you go to your existing cluster and "kubectl edit deployment" the cluster autoscaler and replace the image. This works fine.
> 
> You might have to "kubectl rollout restart" the deployment.

Yeah! that makes sense. Thanks @ghaering

vitobotta:
I will merge this PR and make a release with this change after dinner in a bit. 

@trinhthinh388 If you rerun the create command, hetzner-k3s will automatically update the cluster autoscaler with the new image.

@ghaering If you edit the deployment manifest directly the deployment gets automatically restarted, so there is no need to do it manually.

vitobotta:
Merged. I'll do a quick test and then release.

vitobotta:
Released 2.0.9. Thanks @ghaering for the PR!

PR #513, opened 2025-01-15
docs k3s version fix to latest, other docs-related fixes

MostafaAdly:
First time to see this repo was yesterday, I WAS AMAZED, and i mean it.

now, to the issue in hand :)
the docs fixes had to be made:

- **k3s version mismatch** with the current latest version, i had a problem that every time i try to set it up with the version in docs (v1.23.3+k3s1) it would throw me a a Timeout error that i did not wrap my head around it at all, especially when i've copied everything from the docs and made the needed changes only (1+1 cluster with the types of the nodes)
also it was not related to anything with ssh as i've seen/read the issues/docs and it was not ssh or ssh-agent related stuff

- other logical docs


vitobotta:
Thanks for the fixes!

PR #525, opened 2025-01-30
Fix #434 (subnet sizes != /16)

### LINKED ISSUE #434: Unable to use private networks with subnet size != 16

ValentinVoigt:
Re-created this change due to force-push as discussed in GH-438. I did not test it again.

### DIFF
--- a/src/kubernetes/installer.cr
+++ b/src/kubernetes/installer.cr
@@ -192,7 +192,6 @@ class Kubernetes::Installer
       server: server,
       tls_sans: generate_tls_sans(master_count),
       private_network_enabled: settings.networking.private_network.enabled.to_s,
-      private_network_test_ip: settings.networking.private_network.subnet.split(".")[0..2].join(".") + ".0",
       private_network_subnet: settings.networking.private_network.enabled ? settings.networking.private_network.subnet : "",
       cluster_cidr: settings.networking.cluster_cidr,
       service_cidr: settings.networking.service_cidr,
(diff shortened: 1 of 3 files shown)


vitobotta:
Hi @ValentinVoigt, thanks a lot for remaking the PR! I appreciate it. I just tested it and I really like the solution. Making a new release soon :)

PR #536, opened 2025-02-09
fix: allow udp per default on nodeportrange in firewall

JFWenisch:
Hi!
Per default, a firewall rule `Node port range` is created which allows  `TCP` on the range `30000-32767`. While this totally makes sense to me, I currently don't see any dedicated reason why only TCP should be allowed per default and not UDP. This might be especially relevant as to my knowledge hetzner LB's currently don't support UDP ( see https://github.com/hetznercloud/hcloud-cloud-controller-manager/issues/252 ) .

Thank you for considering this change!

### DIFF
--- a/src/hetzner/firewall/create.cr
+++ b/src/hetzner/firewall/create.cr
@@ -76,9 +76,20 @@ class Hetzner::Firewall::Create
         :destination_ips => [] of String
       },
       {
-        :description => "Node port range",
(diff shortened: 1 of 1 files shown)


vitobotta:
Thanks for submitting the PR! It looks like a safe change, so I've already merged it. It will be included in the next release :)

Issue #541, opened 2025-02-13
Error creating instance: Expected Array or Hash, not Nil


Funzinator:
Hi,
I have the feeling it's a problem with the Hetzner API or something, but I am getting this since yesterday night. I was assuming some sort of rate-limiting because I played around with cluster creation and created many clusters in short time.
However, since it happens still, I wanted to raise this here. Is there a way to see the API responses?

longer log:

``` _          _                            _    _____     
| |__   ___| |_ _____ __   ___ _ __     | | _|___ / ___ 
| '_ \ / _ \ __|_  / '_ \ / _ \ '__|____| |/ / |_ \/ __|
| | | |  __/ |_ / /| | | |  __/ | |_____|   < ___) \__ \
|_| |_|\___|\__/___|_| |_|\___|_|       |_|\_\____/|___/

Version: 2.2.6

[Configuration] Validating configuration...
[Configuration] ...configuration seems valid.
[Placement groups] Deleting unused placement group test-masters...
[Placement groups] ...placement group test-masters deleted
[Placement groups] Creating placement group test-masters...
[Placement groups] ...placement group test-masters created
Error creating instance: Expected Array or Hash, not Nil
Instance creation for test-master1 failed. Try rerunning the create command.
```

And in case it's required, my config:
```
cluster_name: test
kubeconfig_path: "./kubeconfig"
k3s_version: v1.30.3+k3s1

networking:
  ssh:
    port: 22
    use_agent: false
    public_key_path: "/root/.ssh/id_rsa.pub"
    private_key_path: "/root/.ssh/id_rsa"
  allowed_networks:
    ssh:
      - 0.0.0.0/0
    api:
      - 0.0.0.0/0
  public_network:
    ipv4: false
    ipv6: false
  private_network:
    enabled: true
    subnet: 10.63.0.0/16
    existing_network_name: "test"
  cni:
    enabled: true
    encryption: false
    mode: flannel

datastore:
  mode: etcd
  external_datastore_endpoint: postgres://....

schedule_workloads_on_masters: false

image: debian-12
autoscaling_image: debian-12

masters_pool:
  instance_type: cax11
  instance_count: 1
  locations:
    - fsn1

worker_node_pools:
- name: medium-autoscaled
  instance_type: cax11
  location: fsn1
  autoscaling:
    enabled: true
    min_instances: 0
    max_instances: 3

embedded_registry_mirror:
  enabled: false

protect_against_deletion: true

[...]

PR #567, opened 2025-03-24
separate `post_create_commands` for debian and ubuntu

adrobisch:
documenting my findings, see https://github.com/vitobotta/hetzner-k3s/issues/541#issuecomment-2745290642


vitobotta:
Thanks!

PR #569, opened 2025-03-30
fix cloud-init schema validation error when {{ growpart_str }} is not empty

EnDjeee:
{{ growpart_str }}, when not empty, breaks cloud-init schema validation.

This is easily fixed by moving {{ growpart_str }} up in the template file.


vitobotta:
Thanks a lot! Will be part of the next release. :)

PR #571, opened 2025-03-31
Fix output interface in nated setup

madalinignisca:
The output interface for iptables must be the public interface. It can't be the same interface, and usually on hetzner cloud the public one is eth0.

Related to [#541](https://github.com/vitobotta/hetzner-k3s/issues/541)


vitobotta:
Closing, since I think you assumptions are wrong. Hetzner-k3s has had the current configuration for network interfaces for quite a while, and noone has ever reported any issues regarding it.

PR #586, opened 2025-05-01
[WIP] feat: allow to pass robot username & password

Mokto:


Mokto:
I see that you're using https://github.com/hetznercloud/hcloud-cloud-controller-manager/releases/download/v1.23.0/ccm-networks.yaml

We would need to use the helm chart to be able to override the value. Not sure it's easy to do with the current state of the project. Except if you have any other idea ? 

If not I think the PR is ready then.

vitobotta:
Hi, and thanks for the PR. In its current state it seems to only set the credentials for Robot without changing the manifest, and as you point out using the Helm chart would be required for this to work. So it seems the PR is incomplete? If we switch to the Helm chart for the CCM, we need to ensure it can work with existing clusters that were created using the manifest for the CCM.

vitobotta:
The thing is that I was planning to refactor that code so to download the manifest and process it as YAML rather than text, since I have other things I need to do with it. Do you want to take a stab at that?

Mokto:
Done!

vitobotta:
Thanks! This is more like what I had in mind. I will need to test it a bit when I have some time before merging so please bear with me in the meantime.

Mokto:
It worked on public only network but I realized it didn't on private networks.

It's now fixed but it requires a proper CNI system. It works with Cilium, haven't tried the rest: [See explanations here](https://github.com/hetznercloud/hcloud-cloud-controller-manager/blob/v1.24.0/docs/how-to-robot-vswitch-load-balancer.md)

Also updated cloud controller to 1.24 as it improves robot support.

https://github.com/hetznercloud/hcloud-cloud-controller-manager/releases/tag/v1.24.0

src/kubernetes/software/hetzner/cloud_controller_manager.cr
```diff
@@ -25,12 +26,50 @@ class Kubernetes::Software::Hetzner::CloudControllerManager
...
+
+        env_array << robot_enabled
+
+        if settings.networking.private_network.enabled
+          network_routes_enabled = YAML::Any.new({
+            YAML::Any.new("name")  => YAML::Any.new("HCLOUD_NETWORK_ROUTES_ENABLED"),
+            YAML::Any.new("value") => YAML::Any.new("false"),
+          })
+          env_array << network_routes_enabled
+        end
```
Mokto, line 65: support for private networks here

vitobotta:
Thanks for taking this on. It needs some thorough testing, but I'm pretty swamped with work right now, so the merge will have to wait a bit.

Mokto:
Makes sense!

Mokto:
In the meantime added some documentation (that should help you test it) & made sure the CSI driver doesn't run on dedicated nodes.

vitobotta:
Sorry for the delay. I'm still putting this PR on hold because I'm planning to implement a more integrated and native way to add dedicated servers to a cluster with hetzner-k3s. I want it to be more like how we currently do it for other worker nodes. There's a chance I might use some code from your PR, since you've done some work in that area. However, I might close your PR if my implementation differs too much. I’ll keep you updated when I can dive into this properly. Unfortunately, my day job is keeping me really busy right now, so I don’t have much time for other things at the moment.

Mokto:
Thanks for your feedback. I'm not exactly sure what you have in mind but I do think we need to be able to customize dedicated a lot more than what we do with cloud nodes. How we install disks, raid 0, raid 1, etc...

So having that flexibility is quite important in my opinion, especially when you want to add dedicated for databases.

vitobotta:
I understand. I hadn’t really considered how much customization might be needed with dedicated nodes. Let me break down a few key points so I can get a better grasp on your changes:

- Maybe I could add a `show-join-command` option to hetzner-k3s. This would let it print the join command with the token and everything else, so you wouldn’t have to handle that manually. If you check the script currently running on the cloud worker nodes, you'll notice there are some extra settings needed compared to the join command in your changes. Could you update your changes to include those relevant settings?

- I noticed your changes seem to assume that Cilium is the CNI being used. What if someone just goes with the default Flannel? A lot of people stick with the default CNI configuration, which is set up for Flannel.

- Also, why is `K3S_CONTAINERD_SNAPSHOTTER='fuse-overlayfs'` being used?

Thanks!

Mokto:
> * ually. If you check the script currently running on the cloud worker nodes, you'll notice there are some extra settings needed compared to the join command in y

1. Same answer, I don't think the join command should be made from hetzner k3s, because of many reasons. One of the main reasons for me is that I want to pass different labels to it. In my case I'm handling my 50+ dedicated server with ansible.
2. It's just documentation. Just an example. I haven't had the time to check every use case and every CNI. It might just work.
3. This is just documentation, I got the command from somewhere (i think the k3s documentation) and ran it. Not sure what it's doing honestly. Happy to remove it from the documentation

vitobotta:
Thanks! Now, I just need to find the time to thoroughly test this. Unfortunately, my day job is keeping me too busy at the moment. :(

Mokto:
Perfect !!

vitobotta:
Hi @Mokto, I am close to finishing some big refactoring (mostly about improving code quality) that I have been working on for a while. After that, I will start adding your PR. But there might be some conflicts in this PR when I merge my changes. Will you be okay with helping to fix those conflicts? I am sorry this is taking so long.

vitobotta:
Hi @Mokto - it's just some work to make the code better and more organized. I reduced the complexity of some classes by moving parts of the code into separate classes. Things like that. No functionality has changed.

Mokto:
Now I'm the one that is super busy at work 👍

PR #615, opened 2025-07-18
change the order of commands executed in cloud-init runcmd, post_create_commands to be always run first

KristijanL:
This resolves the issue with cluster-autoscaler nodes, as the custom network must be applied first, followed by all other commands.

### DIFF
--- a/src/hetzner/instance/create.cr
+++ b/src/hetzner/instance/create.cr
@@ -301,7 +301,7 @@ class Hetzner::Instance::Create
 
     formatted_additional_commands = format_additional_commands(additional_post_create_commands)
 
-    combined_commands = [post_create_commands, init_commands, formatted_additional_commands].flatten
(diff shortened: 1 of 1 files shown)


vitobotta:
> I have created another branch with additional optimizations that may be beneficial, but I am uncertain whether the project would be interested in incorporating it.
> 
> [cloud-init-test](https://github.com/KristijanL/hetzner-k3s/tree/cloud-init-test)
> 
> During cluster creation testing, I have encountered instances where typos result in inaccessible clusters. To address this issue and ensure the complete removal of resources, I have enhanced the deletion process to prevent failures when the context is unavailable.
> 
> Furthermore, I have modified the logic for the order of post_create_commands to apply only to HCLOUD_CLUSTER_CONFIG.

Can you make a separate PR for the other changes? Thanks!

PR #623, opened 2025-07-28
feat: toggles for CSI-Driver,Traefik,ServiceLB,metrics-server

KaulSe:
This PR makes the installation of several k3s addons configurable from the cluster YAML:
Example:
```yaml
# cluster_config.yaml
csi_driver:
  enabled: false     # skip CSI install

traefik:
  enabled: true      # install Traefik

servicelb:
  enabled: true      # install ServiceLB

metrics_server:
  enabled: true      # install metrics-server
  ```


KaulSe:
Should be fixed :). Thanks!

vitobotta:
Thank you for fixing the conflicts. The PR looks good. I will check if I can merge it and make a new release tomorrow. I was thinking - would it make sense to put these new settings into one group, maybe called `software_components` or something similar? What is your opinion? We could also add a switch for the autoscaler, since someone asked for it in an issue.

KaulSe:
Great idea. I just pushed an update.

- All optional components now live under a single addons: block (addons.csi_driver.enabled, addons.traefik.enabled, …).
- Every addon is its own map, so we can introduce extra per-addon settings later without changing the top-level structure.
- Added a toggle for the Cluster Autoscaler as well (addons.cluster_autoscaler.enabled).
- Docs and sample config updated accordingly.

Let me know if you’d like any further tweaks, but this should make the configuration both cleaner and future-proof.

vitobotta:
I think it's perfect now, and I like the `addons` name for these settings. Thanks! I will let you know once I merge/release.

vitobotta:
Merged! I will do some testing just in case and then make a new release. Thanks again for your contribution!

vitobotta:
Released with v2.3.8.

PR #637, opened 2025-08-04
Moved local path storage class to addons

mavazca:


vitobotta:
> Just adding my two cents as well — we could consider moving the cluster_autoscaler configuration there too. So:
> 
> * `embedded_registry_mirror`
> * `local_path_storage_class`
> * `cluster_autoscaler`
> 
> The only concern I see is that it would break backward compatibility.

As long as the upgrade notes for the new release - the one with these changes - explain clearly how to update the configuration file, it's okay. Would you like to do another PR since you added the addon section? If not, I can handle it for the next release. :)

vitobotta:
Hi! Would you mind adjusting the changes according to the changes in v2.4.0 which I just released? Thanks!

PR #638, opened 2025-08-05
feat: custom hcloud firewall rules

KaulSe:
**Description**

This PR introduces first-class support for arbitrary, granular firewall rules while preserving full backwards-compatibility with existing `ssh` / `api` settings.

### Key Features
1. **Custom firewall rules**
   * New `networking.allowed_networks.custom` section lets users define any number of rules:
     ```yaml
     networking:
       allowed_networks:
         custom:
           - description: "Allow HTTPS from office"
             direction: in
             protocol: tcp
             port: 443
             source_ips:
               - 198.51.100.0/24
           - description: "Egress UDP range"
             direction: out
             protocol: udp
             port: 60000-60100
             destination_ips:
               - 203.0.113.0/24
     ```
   * Supported protocols: `tcp`, `udp`, `icmp`, `esp`, `gre`
   * `direction`: `in` or `out`
   * `port` is validated and only applies to TCP/UDP.
   * Source / destination CIDRs validated (max 100 per rule as per API).

2. **Rule-count safety guard**
   * Hetzner Cloud allows **50 rules per firewall**. `Hetzner::Firewall::Create` now aborts if the generated rule set exceeds this hard limit.

3. **Outbound semantics handled**
   * If **no** `direction: out` rules are present, Hetzner keeps the default *allow-all* egress.
   * As soon as **one** outbound rule is added, egress switches to implicit *deny-all*.  
     This behaviour is documented prominently in the examples.

6. **Compatibility**
   * Existing configs continue to work unchanged.
   * `configure_firewall.sh` (local iptables fallback) remains untouched.

For larger clusters, we should maybe put this feature on the TODO as well (basically doing the same as the hcloud firewall with iptables)?

### DIFF
--- a/docs/Creating_a_cluster.md
+++ b/docs/Creating_a_cluster.md
@@ -38,19 +38,19 @@ networking:
     # in rules (SSH, ICMP, node-port ranges, etc.) use ~10 slots. If the sum of the
     # default rules plus your custom ones exceeds 50, hetzner-k3s will abort with
     # an error.
-    custom_firewall_rules:
-      - description: "Allow HTTP from any IPv4"
-        direction: in
-        protocol: tcp
-        port: 80
-        source_ips:
-          - 0.0.0.0/0
-      - description: "UDP game servers (outbound)"
-        direction: out
-        protocol: udp
-        port: 60000-60100
-        destination_ips:
-          - 203.0.113.0/24
+    # custom_firewall_rules:
+    #   - description: "Allow HTTP from any IPv4"
+    #     direction: in
+    #     protocol: tcp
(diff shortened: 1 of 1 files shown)


KaulSe:
worked for me :)

KaulSe:
Sure :). Changed the code to use `custom_firewall_rules` and changed `source_ips` to an empty array (previously `["0.0.0.0/0"]`)

KaulSe:
alright, makes sense 👍

vitobotta:
Hi! Could you please fix the conflicts and update the changes to match v2.4.0? I just released it. Thanks!

PR #644, opened 2025-08-08
move other addons to addons configuration block

KaulSe:

PR #658, opened 2025-09-10
move other addons to addons configuration block

KaulSe:
new PR (old https://github.com/vitobotta/hetzner-k3s/pull/644)


vitobotta:
Now that I think about it - moving the manifest URLs to the addons section would probably make sense. I mean the settings in `Configuration::Models::Manifests`. Those settings would go into the matching addon sections. Would you like to do this? If not, I'll merge your PR and make those changes before the new release.

vitobotta (approved):

Issue #676, opened 2025-10-30
Creating a cluster with Cilium CNI doesn't work


radekg:
I'm using the following configuration file:

```yaml
---
kubeconfig_path: "~/.hetzner/kube/gmbh"
cluster_name: gmbh-infrastructure
k3s_version: v1.30.3+k3s1

networking:
  ssh:
    port: 22
    use_agent: false
    public_key_path: "~/.hetzner/ssh/gmbh_rsa.pub"
    private_key_path: "~/.hetzner/ssh/gmbh_rsa"
  allowed_networks:
    ssh: 
    - 0.0.0.0/0
    api:
    - 0.0.0.0/0
  public_network:
    ipv4: true
    ipv6: false
  private_network:
    enabled: true
    subnet: 10.0.0.0/16
  cni:
    enabled: true
    encryption: false
    mode: cilium
    cilium:
      # Optional: specify a path to a custom values file for Cilium Helm chart
      # When specified, this file will be used instead of the default values
      helm_values_path: "/home/radek/dev/hetzner/gmbh/cilium-values.yaml"
      chart_version: "v1.18.2"
  cluster_cidr: 10.244.0.0/16 # optional: a custom IPv4/IPv6 network CIDR to use for pod IPs
  service_cidr: 10.43.0.0/16 # optional: a custom IPv4/IPv6 network CIDR to use for service IPs. Warning, if you change this, you should also change cluster_dns!
  cluster_dns: 10.43.0.10 # optional: IPv4 Cluster IP for coredns service. Needs to be an address from the service_cidr range

schedule_workloads_on_masters: false
protect_against_deletion: false

# image: rocky-9 # optional: default is ubuntu-24.04
# snapshot_os: microos # otional: specified the os type when using a custom snapshot
masters_pool:
  instance_type: cpx21
  instance_count: 1
  locations:
  - nbg1

worker_node_pools:
- name: small
  instance_type: cpx21
  instance_count: 3
  location: nbg1
```

The cluster never comes up successfully:

```
[Configuration] Validating configuration...
[Configuration] ...configuration seems valid.
[Private Network] Creating private network...
[Private Network] ...private network created
[SSH key] Creating SSH key...
[SSH key] ...SSH key created
[Instance gmbh-infrastructure-master1] Creating instance gmbh-infrastructure-master1 (attempt 1)...
[Instance gmbh-infrastructure-master1] Instance status: initializing
[...]

PR #677, opened 2025-11-05
Install software after workers

radekg:
This PR addresses #676. Install the software after workers so that when Cilium is installed, there are workers to deploy the CNI pods to.

### DIFF
--- a/src/kubernetes/installer.cr
+++ b/src/kubernetes/installer.cr
@@ -43,7 +43,7 @@ class Kubernetes::Installer
     @configuration,
     @load_balancer,
     @ssh,
-    @autoscaling_worker_node_pools
+    @autoscaling_worker_node_pools,
   )
     @kubeconfig_manager = Kubernetes::KubeconfigManager.new(@configuration, settings, @ssh)
     @master_generator = Kubernetes::Script::MasterGenerator.new(@configuration, settings)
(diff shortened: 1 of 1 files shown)


vitobotta:
Hi @radekg , sorry for the delay. I cannot merge the pull request as is because it breaks the creation of the cluster for default and more common configurations. By moving the installation of the software to after the installation of k3s on the workers, the `create` command gets stuck after setting up the workers because they cannot reach a ready state without the CCM being installed first. And then the Cluster Autoscaler installation has to wait for at least one worker to be ready. So the process hangs. Can you please try to figure out a way to make it work with your Cilium config without affecting more common use cases? As another user pointed out in the original thread, they don't have problems with Cilium on several clusters, so it has to be an issue with your particular configuration that requires some changes.

vitobotta (changes requested):
See other comment :)

codepainters:
@radekg can it be due to the taints problem I've reported as #744 ? Cilium doesn't need worker nodes to progress with installation.

vitobotta:
Hi @radekg  - any update on this?

radekg:
Hey @vitobotta, I'm sorry, I do not have any update on this. Feel free to close the ticket if there are ways around.

vitobotta:
Ok, thanks

PR #683, opened 2025-11-22
Add private_registry_config to embedded_registry_mirror.cr

### LINKED ISSUE #682: How to create my own registires.yaml
Install script [writes](https://github.com/vitobotta/hetzner-k3s/blob/main/templates/master_install_script.sh#L106) registries.yaml as default.

I found out that pre and post commands are executed before install script and it overwrite my registries.yaml.

Do I understand everything right? What options do I have without applying a patch to hetzner-k3s?

For me now it seems that I have to create a PR where content for registries is taken from config.

sergioboi:
It creates a way to generate custom /etc/rancher/k3s/registries.yaml to modify images pulling logic https://docs.k3s.io/installation/private-registry
Resolves #682


sergioboi:
I've just compiled the sources. If changes seem ok, I could try to deploy to my hetzner sandbox cluster to be sure that everything works.

src/configuration/models/embedded_registry_mirror.cr
```diff
@@ -3,6 +3,7 @@ class Configuration::Models::EmbeddedRegistryMirror
   include YAML::Serializable::Unmapped
 
   getter enabled : Bool = true
+  getter private_registry_config : String? = nil
```
vitobotta: There should be a default value here matching the current behavior, not nil.

vitobotta (changes requested):
Looks good but we need to tweak a couple of things :)

vitobotta:
Thanks for the PR! See other comments :)

sergioboi:
@vitobotta thank you for the comments. I've changed everything you mentioned. Also, I've tested it on my clusters, so default value works.

vitobotta:
Hi @sergioboi , it seems that after merging previous PRs there are a couple of conflicts in this one. Do you mind fixing them so I can merge it? Thanks!

vitobotta:
> > Hi @sergioboi , it seems that after merging previous PRs there are a couple of conflicts in this one. Do you mind fixing them so I can merge it? Thanks!
> 
> I gave a look at changes you've made. And I have a question where better to put registries config. I'm not sure that addons is appropriate place. What do you think?

There's a new section in the config for the embedded registry mirror, so ideally it should go there. :)

sergioboi:
> > > Hi @sergioboi , it seems that after merging previous PRs there are a couple of conflicts in this one. Do you mind fixing them so I can merge it? Thanks!
> > 
> > 
> > I gave a look at changes you've made. And I have a question where better to put registries config. I'm not sure that addons is appropriate place. What do you think?
> 
> There's a new section in the config for the embedded registry mirror, so ideally it should go there. :)

Done

vitobotta (approved):
Thanks! Code wise it looks good. I'll do some quick tests one of these days. and let you know when I merge it. :)

PR #672, opened 2025-10-17
Add pagination support for Server Types API

artem-zinnatullin:
On Oct 15 2025 Hetzner announced restructurization of their pricing plans and introduced bunch of new server / instance types. 

hetzner-k3s can't create pools of some instance types now like `cx33`:

```js
[Configuration] Validating configuration...
[Configuration] Some information in the configuration file requires your attention, aborting.
[Configuration]  - masters node pool has an invalid instance type
```

See issue #671.

The API for resolving server types now returns 2 pages of information:

```bash
curl \
  -H "Authorization: Bearer $API_TOKEN" \
  "https://api.hetzner.cloud/v1/server_types"
```

API Docs: https://docs.hetzner.cloud/reference/cloud#server-types

It's now a paginated response:

```json
"meta": {
  "pagination": {
   "last_page": 2,
   "next_page": 2,
   "page": 1,
   "per_page": 25,
   "previous_page": null,
   "total_entries": 29
  }
```

This PR adds support for iterating all pages of Server Type API response and collecting instance types from that.

Tested locally via:

```bash
brew install crystal
shards install --without-development
crystal build src/hetzner-k3s.cr --release
chmod +x hetzner-k3s

hetzner-k3s create --config test-cluster-with-instance_type-cx33.yaml
 _          _                            _    _____
| |__   ___| |_ _____ __   ___ _ __     | | _|___ / ___
| '_ \ / _ \ __|_  / '_ \ / _ \ '__|____| |/ / |_ \/ __|
| | | |  __/ |_ / /| | | |  __/ | |_____|   < ___) \__ \
|_| |_|\___|\__/___|_| |_|\___|_|       |_|\_\____/|___/

Version: 2.4.2

[Configuration] Validating configuration...
[Configuration] ...configuration seems valid.
[SSH key] Creating SSH key...
[SSH key] ...SSH key created
[Instance xxx-master1] Creating instance xxx-master1 (attempt 1)...
[Instance xxx-master1] Instance status: initializing
[Instance xxx-master1] Powering on instance (attempt 1)
[Instance xxx-master1] Waiting for instance to be powered on...
…
[System Upgrade Controller] ...System Upgrade Controller installed
[Control plane] Switched to context "xxx-master1".
```

Apologies if my Crystal code isn't very idiomatic, first time I did anything in this language 😅 

Closes #671.

PR #689, opened 2025-11-27
Bump outdated version number in docs

lenn4rd:
I copy-pasted the installation instructions which use 2.4.1. This version doesn't support pagination so the tool is unaware of recently added server types (see #672). Only after diving into the source I code I discovered I'm running an outdated version.

Perhaps we can automate updating installation instructions by using placeholders in the Markdown or search/replace in the release workflow. Let me know what would be your preferred approach, happy to contribute.


vitobotta:
Thanks for spotting this!

PR #697, opened 2025-12-19
feat(hetzner-k3s): HCLOUD_CLUSTER_CONFIG_FILE 

leblancmeneses:
related to: https://github.com/vitobotta/hetzner-k3s/discussions/696

solution is documented in cluster-autoscaler: https://github.com/kubernetes/autoscaler/blob/master/cluster-autoscaler/cloudprovider/hetzner/README.md

[quote]
HCLOUD_CLUSTER_CONFIG_FILE Can be used as alternative to HCLOUD_CLUSTER_CONFIG.
[/quote]

Switched binary to use HCLOUD_CLUSTER_CONFIG_FILE internally instead.

This is working in my cluster with over 20+ configured node pools.

I tried running the e2e tests but they are not working on main for me after setting up env so it won't work on my branch.
```
[Configuration] Some information in the configuration file requires your attention, aborting.
[Configuration]  - The number of locations specified for masters must equal the total number of masters
[Configuration]  - All workers must be in valid locations and in the same network zone as the masters when using a private network or etcd datastore
 _          _     
```

### DIFF
--- a/.claude/settings.local.json
+++ b/.claude/settings.local.json
@@ -3,9 +3,7 @@
     "allow": [
       "Bash(crystal tool format:*)",
       "Bash(grep:*)",
-      "Bash(crystal build:*)",
-      "Bash(./hetzner-k3s create:*)",
-      "Bash(tee:*)"
+      "Bash(crystal build:*)"
     ],
     "deny": []
   }


vitobotta:
Thanks for the PR! And sorry for the long delay. I haven't had much free time in ages due to the day job. I will make a new release today.

PR #699, opened 2025-12-23
Fix grow_root_partition_automatically for ClusterAutoscaler nodes

### LINKED ISSUE #698: Global grow_root_partition_automatically: false is not respected for nodes created by ClusterAutoscaler
Hi @vitobotta!

I believe I found a major bug in cloud-init for nodes created by ClusterAutoscaler.

v2.4.3 hetzner-k3s.yaml like this:

```yaml
grow_root_partition_automatically: false

additional_pre_k3s_commands:
  - sgdisk -e /dev/sda
  - partprobe
  - parted -s /dev/sda mkpart mydatapartition ext4 20GB "100%"
  - growpart /dev/sda 1
  - resize2fs /dev/sda1
  - partprobe
  - udevadm settle

addons:
  cluster_autoscaler:
    enabled: true
```

Produces following behavior:

- Nodes created via static pools behave correctly: `/etc/growroot-disabled` is present, root partition stays at default 4GB and remaining space is available for StorageClass customization ie, LUKS, OpenEBS, etc via additional_pre_k3s_commands / additional_post_k3s_commands, additional_pre_k3s_commands are written to `/var/lib/cloud/instance/user-data.txt` as well as `write_files: path: /etc/growroot-disabled`
- Nodes created by ClusterAutoscaler however are broken, `/etc/growroot-disabled` is not created, `/var/lib/cloud/instance/user-data.txt` contains `additional_pre_k3s_commands` but doesn't have the `write_files: path: /etc/growroot-disabled` section! 

As a result nodes created by `ClusterAutoscaler` don't respect the global `grow_root_partition_automatically: false` and automatically consume entire NVMe for the root partition making them unusable for custom StorageClass solutions.

I believe the issue is caused by the use of default parameter value`@grow_root_partition_automatically = true` in multiple places across the code:

- https://github.com/vitobotta/hetzner-k3s/blob/55a288df3df60d09d5f26ded9630d41322ba98f6/src/hetzner/instance/create.cr#L63
- https://github.com/vitobotta/hetzner-k3s/blob/55a288df3df60d09d5f26ded9630d41322ba98f6/src/hetzner/instance/create.cr#L84

I think the value should be set explicitly and default to true only in one place in the code instead of having a default value in function/constructor definitions?

artem-zinnatullin:
With this change nodes created by ClusterAutoscaler now respect global `grow_root_partition_automatically: false` as expected:

```yaml
grow_root_partition_automatically: false

addons:
  cluster_autoscaler:
    enabled: true
```

Checked on a node created by ClusterAutoscaler:

```js
root@tommy-test-cx23-workers-pool-xxx:~# cat /var/lib/cloud/instance/user-data.txt
ls -la /etc/growroot-disabled
-rw-r--r-- 1 root root 5 Dec 23 02:16 /etc/growroot-disabled

root@tommy-test-cx23-workers-pool-xxx:~# cat /var/lib/cloud/instance/user-data.txt
#cloud-config
preserve_hostname: true



write_files:
- content: |
    true
  path: /etc/growroot-disabled
```

Fixes #698.


privatecoder:
@artem-zinnatullin check [this](https://github.com/vitobotta/hetzner-k3s/compare/main...privatecoder:hetzner-k3s:fix-autoscaler-grow-root-partition) (you have that change as well).

@vitobotta: For autoscaled nodes the `grow_root_partition_automatically`-setting isn't passed to `cloud_init`, so it falls back to `true`, which skips creating `/etc/growroot-disabled`.

I have just tested it with only thes two lines added and it works fine, i.e. `/etc/growroot-disabled` is created on an autoscaled node with the global setting `grow_root_partition_automatically: false`

vitobotta:
Closing in favour of https://github.com/vitobotta/hetzner-k3s/pull/735.

PR #730, opened 2026-02-06
Add ssh key check for autoscaler

privatecoder:
#### Attempted goal
1. Set up a dev cluster with autoscaling nodes
2. Be able to ssh into autoscaled nodes

#### Issues
• Could not ssh into autoscaled nodes, because the ssh key was not added to `authorized_keys` on the node
• After investigating I found that yout tool sets `HCLOUD_SSH_KEY` in the cluster‑autoscaler deployment to
  `settings.cluster_name` so the autoscaler injects that key when it creates nodes.
• As I added the key before under a slightly different name, this did not work

#### Solution
Added a check, so that if any worker pool has `autoscaling.enabled`: `true` and `addons.cluster_autoscaler.enabled`: `true`, it:
- Finds the SSH key by fingerprint or name.
- If the key found has a different name than cluster_name, it aborts with a clear error that autoscaler won’t work.
- If no key exists yet, it lets create proceed (hetzner‑k3s will create it).

If it fails, it prints:

`[Configuration]  - Cluster autoscaler requires an SSH key named 'dev' in Hetzner. A key with the same fingerprint exists as 'dev1', so hetzner-k3s will not create 'dev'. Autoscaled nodes will be created without SSH keys. Rename or delete the existing key, or change cluster_name.`

### DIFF
--- a/src/configuration/validators/create_settings.cr
+++ b/src/configuration/validators/create_settings.cr
@@ -12,6 +12,7 @@ require "./masters_pool"
 require "./worker_node_pools"
 require "./kubectl_presence"
 require "./helm_presence"
+require "../../hetzner/ssh_key/find"
 
 class Configuration::Validators::CreateSettings
   getter errors : Array(String) = [] of String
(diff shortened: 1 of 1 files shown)


vitobotta:
Can you please open a new PR for this specific issue? This PR contains many unrelated changes :)

Issue #474, opened 2024-10-27
Limit Hetzner firewall port exposure


roy-hardin:
Wouldn’t security be improved if the `hetzner-k3s create` command of hetzner-k3s version 2.0.8 did not (re-)open the Hetzner Cloud Firewall to allow any incoming traffic from any IPv4 or IPv6 address on ports 5001, 30000 through 32767, and for ICMP (ping)?

According to the [OWASP Kubernetes Security Cheat Sheet](https://cheatsheetseries.owasp.org/cheatsheets/Kubernetes_Security_Cheat_Sheet.html#controlling-network-access-to-sensitive-ports):

_Since Kubernetes clusters usually listen on a range of well-defined and distinctive ports, it is easier for attackers to identify the clusters and attack them. An overview of the default ports used in Kubernetes is provided below. Make sure that your network blocks access to ports, and you should seriously consider limiting access to the Kubernetes API server to trusted networks._

_Control plane nodes_

Protocol | Port Range | Purpose
-- | -- | -- 
TCP | 6443- | Kubernetes API Server 
TCP | 2379-2380 | etcd server client API 
TCP | 10250 | Kubelet API
TCP | 10251 | kube-scheduler
TCP | 10252 | kube-controller-manager 
TCP | 10255 | Read-Only Kubelet API 

_Worker Nodes_

Protocol | Port Range | Purpose
-- | -- | --
TCP | 10250 | Kubelet API
TCP | 10255 | Read-Only Kubelet API
TCP | 30000-32767 | NodePort Services

Given these guidelines, would it be possible to restrict the firewall configuration to enhance security while preserving necessary functionality?

roy-hardin:
Thanks for your detailed response. I understand the port requirements for 5001 and 30000-32767 and appreciate your explanation about the configuration. It’s reassuring that peer connections are authenticated, and NodePort services only pose a risk if exposed. Configuring the NodePort range and ICMP settings in the future would be great for those who want more control over the firewall setup.

vitobotta:
Hi @roy-hardin , I guess we can close this?

PR #732, opened 2026-02-08
make NodePort range configurable and add ability to disable it alltogether

privatecoder:
I think @roy-hardin had a valid point in #474. Even though you wrote _"Ports 30000-32767 are required for NodePort services, so they only pose a risk if you explicitly decide to expose a service to the Internet on a NodePort. Otherwise those ports are not in use so they don't pose any risk IMO."_ I think that opening the full NodePort range to source:any leaves an unnecessary attack surface.

Because: Even if we don’t plan to use NodePorts directly, sometimes they can appear implicitly (chart defaults, upgrades, misclicks) and would become publicly reachable immediately. Having an option to disable them or make them configurable keeps the flexibility while avoiding accidental exposure.

Small off-topic question: Did I get it right and if so: is it intentional to not use `custom_firewall_rules` for the custom firewall when using `use_local_firewall`: `true`?

### DIFF
--- a/docs/Creating_a_cluster.md
+++ b/docs/Creating_a_cluster.md
@@ -57,6 +57,8 @@ networking:
     #     port: 60000-60100
     #     destination_ips:
     #       - 203.0.113.0/24
+  # node_port_firewall_enabled: true # optional: set false to disable NodePort firewall rules (TCP/UDP)
+  # node_port_range: "30000-32767" # optional: NodePort range to open on firewalls (TCP/UDP)
   public_network:
     ipv4: true
     ipv6: true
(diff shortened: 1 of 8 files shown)


vitobotta:
> > One thing confuses me in the PR. You removed the firewall rules for the NodePort range, but you did not add anything to replace them when the Hetzner firewall is used
> 
> I don't understand that – I did not remove them, but made them conditional (`if node_port_firewall_enabled`). Or what do you mean?

I wonder if it was a Github glitch because I see the rules now. Weird.

vitobotta:
Thanks! Will make a release soon.

PR #758, opened 2026-04-27
add cluster_domain configuration parameter

an-tex:


vitobotta (approved):

vitobotta:
Looks good, and but a thought came to my mind after approving it. Can you please omit specifying the argument if the cluster domain is left to the default value? Otherwise this change will cause a restart of k3s on all masters for existing clients when re-running the `create` command. Thanks

src/kubernetes/script/master_generator.cr
```diff
@@ -48,6 +48,7 @@ class Kubernetes::Script::MasterGenerator
       cluster_cidr:                     @settings.networking.cluster_cidr,
       service_cidr:                     @settings.networking.service_cidr,
       cluster_dns:                      @settings.networking.cluster_dns,
+      cluster_domain:                   @settings.networking.cluster_domain || "",
```
an-tex, line 51: not sure if that's the preferred way. but without the ` || ""`, the value "none" would be passed to the template

an-tex:
> Looks good, and but a thought came to my mind after approving it. Can you please omit specifying the argument if the cluster domain is left to the default value? Otherwise this change will cause a restart of k3s on all masters for existing clients when re-running the `create` command. Thanks

good point! I've marked the cluster_domain as optional and only add the `k3s server --cluster-domain=...` parameter if it's defined.

tested both cases using dev containers with my cluster

vitobotta:
Thanks for the contribution! I will make a new release soon.

PR #768, opened 2026-05-11
feat: tailscale integration

dsbibby:
This PR adds optional [Tailscale](https://tailscale.com) integration to k3s nodes, allowing private access to nodes via a tailnet.

It also enables IPv6-only nodes by including support for [464XLAT](https://datatracker.ietf.org/doc/html/rfc6877), which provides IPv4 connectivity on IPv6-only Hetzner instances.

I built and tested this for my own small cluster, but then wondered if the project would like it as a contribution. No hard feelings if it's not a good fit for core hetzner-k3s — happy to share it either way.

**What's changed:**
- Tailscale is optionally installed and joined on each node at provision time via new config options
- SSH connections are routed via Tailscale for IPv6-only nodes (which have no public IPv4)
- 464XLAT (clatd) is configured on IPv6-only nodes to maintain IPv4 compatibility for k3s and workloads

**Testing:**
Three end-to-end tests were run against real Hetzner infrastructure (single master + single worker, `cax11`, `nbg1`):
- IPv4, no Tailscale
- IPv4 + Tailscale
- IPv6-only + Tailscale

All three passed with nodes Ready and pods Running.

Finally, I have a multi master cluster deployed as IPv6 only with tailscale and running production workloads.

Comments, questions and suggestions very much welcomed. Thank you for this great tool!


src/configuration/validators/networking_config/allowed_networks.cr
```diff
@@ -8,8 +8,9 @@ class Configuration::Validators::NetworkingConfig::AllowedNetworks
   getter errors : Array(String) = [] of String
   getter allowed_networks : Configuration::Models::NetworkingConfig::AllowedNetworks
   getter skip_current_ip_validation : Bool = false
+  getter use_tailscale : Bool = false
```
vitobotta, line 11: Why is this setting in allowed_networks.cr and not in the SSH class where it's more appropriate?

src/hetzner/instance/cloud_init_generator.cr
```diff
@@ -191,12 +247,31 @@ class Hetzner::Instance::CloudInitGenerator
   end
 
   private def mandatory_post_create_commands
-    commands = [
-      "hostnamectl set-hostname $(curl http://169.254.169.254/hetzner/v1/metadata/hostname)",
-      "update-crypto-policies --set DEFAULT:SHA1 || true",
+    commands = [] of String
+
+    commands << "/usr/local/bin/configure-dns.sh" unless @settings.networking.dns_servers.empty?
```
vitobotta, line 252: Why have you removed the `"update-crypto-policies --set DEFAULT:SHA1 || true"` line?

src/util/ssh.cr
```diff
@@ -66,10 +82,6 @@ class Util::SSH
           log_line "Instance #{instance.name} not ready, retrying...", log_prefix: "Instance #{instance.name}"
           raise IO::Error.new("Result mismatch")
         end
```
vitobotta, line 84: I prefer to keep a timeout restriction in case ssh hangs.

docs/Creating_a_cluster.md
```diff
@@ -23,6 +23,7 @@ networking:
     public_key_path: "~/.ssh/id_ed25519.pub"
     private_key_path: "~/.ssh/id_ed25519"
     # existing_ssh_key_name: "my-existing-key" # optional: use an existing SSH key in Hetzner instead of creating a new one
+    # ssh_wait_attempts: 20 # how many SSH attempts to make while waiting for a node (5s between each)
```
vitobotta, line 26: This setting doesn't seem to be used anywhere. Did you forget to implement it?

src/configuration/models/networking.cr
```diff
@@ -20,6 +20,7 @@ module Configuration
       getter cluster_cidr : String = "10.244.0.0/16"
       getter service_cidr : String = "10.43.0.0/16"
       getter cluster_dns : String = "10.43.0.10"
+      getter dns_servers : Array(String) = [] of String
```
vitobotta, line 23: Since you have removed the prior DNS resolver config, I'd default this array to using the Cloudflare DNS resolvers.

src/hetzner/instance/cloud_init_generator.cr
vitobotta, line 252: Since we are defaulting to Cloudflare DNS, we can assume the setting will never be empty. We can actually add a validation in case by mistake the user sets the setting to an empty array.

templates/configure_dns.sh
```diff
@@ -0,0 +1,39 @@
+#!/usr/bin/env python3
```
vitobotta, line 1: Make sure python is available by default in all the official images available in Hetzner Cloud. If not, we either need to detect the OS and install python when needed, or switch to regular bash.

vitobotta (changes requested):
Overall this is one of the best PRs I have received so far. Thanks for the contribution! Please see comments.

src/configuration/validators/networking_config/allowed_networks.cr
dsbibby: Good point and on reflection, I agree `AllowedNetworks` shouldn’t need to know about Tailscale specifically. A cleaner version would keep that decision in the higher-level networking validator and pass it through as `skip_current_ip_validation`, e.g. `skip_current_ip_validation || networking.ssh.use_tailscale`.

I don’t think this belongs in the runtime SSH helper, since this is configuration validation rather than SSH connection behavior, but I agree the Tailscale-specific flag leaking into `allowed_networks.cr` is unnecessary. I’ll clean that up.

src/hetzner/instance/cloud_init_generator.cr
dsbibby: Good spot! This was unintentionally removed during my development of this. I'll restore it.

src/util/ssh.cr
dsbibby: I'm still learning crystal, but I think these rescue/raise lines are no-ops as they are just re-raising the `Tasker::Timeout` exception without any additional handling? 

The timeout restriction is still there via the surrounding `Tasker.timeout(...) do` block above.

docs/Creating_a_cluster.md
dsbibby: It is implemented in `src/hetzner/instance/create.cr`:

```crystal
ssh_client.wait_for_instance instance, ssh.port, ssh.use_agent, "echo ready", "ready", ssh.ssh_wait_attempts 
```

src/configuration/models/networking.cr
dsbibby: The previous kubelet resolver setup was moved into the master/worker install scripts, but `networking.dns_servers` currently only configures the host upstream resolvers via net plan which I think actually is confusing for users. 

Additionally IPv6-only nodes need their own sensible default because the previous `8.8.8.8` resolver is not reachable without IPv4.

I think the behavior should be:
- IPv4 nodes default to the previous `8.8.8.8` resolver unless `dns_servers` is configured
- IPv6-only nodes default to IPv6/DNS64-capable resolvers unless `dns_servers` is configured
- IPv6-only kubelet/CoreDNS should still point at the node-local `systemd-resolved` proxy, because pods use the IPv4 pod network and need an IPv4-reachable DNS endpoint

I’ll update this so `dns_servers` consistently controls upstream DNS, with protocol-appropriate defaults.

PR #771, opened 2026-05-14
Add configurable gVisor (runsc) container runtime support

### LINKED ISSUE #770: Support configuring gVisor (runsc) as an additional container runtime
## Problem

Currently, hetzner-k3s only supports the default containerd (runc) runtime. Users who want to run workloads with gVisor for additional sandboxing have to manually SSH into each node to install runsc and configure containerd.

## Proposed Solution

Add a configurable gVisor addon that:

1. Installs the `runsc` package from the official gVisor apt repo on all nodes during cluster creation
2. Creates the containerd `config.toml.tmpl` file at `/var/lib/rancher/k3s/agent/etc/containerd/config.toml.tmpl` with the runsc runtime configuration
3. Applies a Kubernetes `RuntimeClass` resource so workloads can use `runtimeClassName: gvisor`

The feature is opt-in via the config file:

```yaml
addons:
  gvisor:
    enabled: true
```

## Implementation Notes

- Uses the `io.containerd.cri.v1.runtime` plugin path (compatible with recent containerd versions shipped with k3s)
- The containerd config template uses `{{ template base . }}` to preserve k3s's default containerd configuration
- RuntimeClass `handler: runsc` is applied server-side after cluster creation
- Follows the guide at https://gist.github.com/Frichetten/c77ee24b12edd2ab852738fc8221a1f1

rightsum:
Closes #770

## What

Adds opt-in gVisor support as an additional container runtime alongside the default containerd/runc. Managed entirely through the cluster config file.

## Changes

- New `Configuration::Models::AddonsConfig::GVisor` config model
- New `Kubernetes::Software::GVisor` installer (applies RuntimeClass)
- Modified master+worker install script templates to install runsc and configure containerd when gVisor is enabled
- Modified script generators to pass gvisor_enabled flag to templates

## Usage

```yaml
addons:
  gvisor:
    enabled: true
```

After cluster creation, pods can use:
```yaml
spec:
  runtimeClassName: gvisor
```

## Verification

- [x] `crystal build --no-codegen` passes
- [x] `crystal tool format --check` passes
- [x] Follows existing code patterns (addons config + software installer)

### DIFF
--- a/src/kubernetes/script/master_generator.cr
+++ b/src/kubernetes/script/master_generator.cr
@@ -53,11 +53,12 @@ class Kubernetes::Script::MasterGenerator
       embedded_registry_mirror_enabled: @settings.addons.embedded_registry_mirror.enabled.to_s,
       private_registry_config:          @settings.addons.embedded_registry_mirror.private_registry_config,
       local_path_storage_class_enabled: @settings.addons.local_path_storage_class.enabled.to_s,
-      traefik_enabled:        @settings.addons.traefik.enabled.to_s,
-      servicelb_enabled:      @settings.addons.servicelb.enabled.to_s,
-      metrics_server_enabled: @settings.addons.metrics_server.enabled.to_s,
-      labels_and_taints:      labels_and_taints,
-      additional_post_k3s_commands: post_k3s_commands,
+      traefik_enabled:                  @settings.addons.traefik.enabled.to_s,
+      servicelb_enabled:                @settings.addons.servicelb.enabled.to_s,
+      metrics_server_enabled:           @settings.addons.metrics_server.enabled.to_s,
+      labels_and_taints:                labels_and_taints,
+      gvisor_enabled:                   @settings.addons.gvisor.enabled?.to_s,
+      additional_post_k3s_commands:     post_k3s_commands,
     })
   end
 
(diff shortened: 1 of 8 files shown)


src/kubernetes/script/master_generator.cr
```diff
@@ -53,11 +53,12 @@ class Kubernetes::Script::MasterGenerator
...
-      traefik_enabled:        @settings.addons.traefik.enabled.to_s,
-      servicelb_enabled:      @settings.addons.servicelb.enabled.to_s,
-      metrics_server_enabled: @settings.addons.metrics_server.enabled.to_s,
-      labels_and_taints:      labels_and_taints,
-      additional_post_k3s_commands: post_k3s_commands,
+      traefik_enabled:                  @settings.addons.traefik.enabled.to_s,
+      servicelb_enabled:                @settings.addons.servicelb.enabled.to_s,
+      metrics_server_enabled:           @settings.addons.metrics_server.enabled.to_s,
+      labels_and_taints:                labels_and_taints,
+      gvisor_enabled:                   @settings.addons.gvisor.enabled?.to_s,
```
vitobotta, line 60: Nit: Move this line to before the previous so the booleans for the addons are all together.

vitobotta (changes requested):
Very good PR, thanks! Please see my comments. :)

PR #781, opened 2026-06-19
Add Cilium network validation checks, MTU settings

Yiidiir:
## Summary

Adds MTU and connectivity validation to the Cilium installation flow to catch broken network configurations before they reach production.

## Changes

- Adds `networking.cni.cilium.mtu` and `underlay_mtu` configuration fields.
- Validates Cilium MTU settings during configuration loading.
- Runs a pre-deployment MTU validation before `helm upgrade --install`.
- Reads MTU from custom Cilium Helm values when `helm_values_path` is used.
- Adds post-rollout cross-node network validation after `kubectl rollout status ds cilium`.
- Gates installation success on both:
  - small packet pod-to-pod ping
  - large 1230B pod-to-pod ping with DF set

## Why

This prevents Cilium installs from succeeding when the configured MTU does not fit the network path MTU after VXLAN encapsulation overhead. It also avoids relying only on rollout status or small pings, which can pass while larger packets are still blackholed.

## Validation

- `git diff --check`
- `crystal build src/hetzner-k3s.cr --no-codegen` using Crystal 1.20.0 Docker image

The build passes with only existing `Time.monotonic` deprecation warnings.

### DIFF
--- a/docs/Creating_a_cluster.md
+++ b/docs/Creating_a_cluster.md
@@ -80,6 +80,8 @@ networking:
       # When specified, this file will be used instead of the default values
       # helm_values_path: "./cilium-values.yaml"
       # chart_version: "v1.17.2"
+      # mtu: 1450 # optional: Cilium underlay MTU; pod MTU is reduced by VXLAN overhead
+      # underlay_mtu: 1450 # optional: validation-only network path MTU constraint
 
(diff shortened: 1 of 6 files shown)


Yiidiir:
@vitobotta this addresses issues with MTU size when trying to run tailscale for certain pods inside the cluster

